PR #6, opened 2015-01-02
Count the frames in all recordings of a directory

kimtonyhyun:
@forea

Running `count_frames_in_xml(pwd)` in a directory that contains Miniscope recordings (i.e. XML files) will aggregate the total number of frames indicated in all XML files. This function allows us to quickly assess the number of frames in a recording session, given that each trial is now a separate recording.

As courtesy, `count_frames_in_xml` will output a warning if it is unable to find a RAW or TIF file that corresponds to the XML file.

`parse_miniscope_xml` is a helper function that will parse (some of) the content of the Miniscope XML file into a format (struct) easily parsed in Matlab.

### DIFF
--- a/io/count_frames_in_tif.m
+++ b/io/count_frames_in_tif.m
@@ -1,0 +1,26 @@
+function total_frames = count_frames_in_tif(path_to_tif)
+% Computes the number of frames contained in all TIF files of
+%   the specified directory
+%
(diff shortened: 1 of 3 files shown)


kimtonyhyun:
@forea @jmaxey 

Also added a function to count the frames in all TIF files (i.e. to be run after decompression)

PR #7, opened 2015-01-08
added make_summary.m, which creates a mat file with the summary of multiday data

inanhkn:
@kimtonyhyun @forea @jmaxey 
The function make_summary takes a struct 'sources' as input (examine an example 'data_sources.m' for the organization of the struct). 

it creates a mat file with the summary as explained in the documentation below. 

Notes:
1) it is assumed that the behavior video has  640x480 resolution.
2) compression of the behavior video frame indices is done by appending 100 frames before and after a gate action (should probably be fed as an input or be modified for the current cohort together with some other functions that read maze data).


inanhkn:
@kimtonyhyun @forea @jmaxey
I also added an example data_sources.m file that creates an input struct. assuming the files exist in the specified directories, make_summary(data_sources) should work.

PR #8, opened 2015-01-08
extract the xy coordinates of the mouse throughout the behavior video

inanhkn:
@kimtonyhyun @jmaxey @forea 
Extract X and Y coordinates of the mouse from the behavior video

The input video is that of a single trial. It is input together with a snapshot of the behavior scene to be used for background subtraction. After the background subtraction the image (-hopefully- biggest object in which is the mouse) is binarized and blobs are extracted. Mouse X Y coordinates are simply the center of mass of the largest blob.

### DIFF
--- a/behavior/get_spatial_trace_from_behavior.m
+++ b/behavior/get_spatial_trace_from_behavior.m
@@ -1,0 +1,62 @@
+function spatial_trace = get_spatial_trace_from_behavior(sample_image,video)   
+%function spatial_trace = get_spatial_trace_from_behavior(sample_image,video)
+%
+%Extract X and Y coordinates of the mouse from the behavior video
+%
+%   sample_image : A snapshot the behavior setup for use in the maze region
+%   extraction step.
+%
+%   video : [Height x Width x num_frames] array
+%
+%   spatial_trace : [num_frames x 2] matrix where the 1st column is the X-coordinate and
+%   the 2nd column is the Y-coordinate of the mouse. 
+
+    
+    num_frames = size(video,3);
+    centroids = zeros(num_frames,2); 
+    
+    %anything above thresh_whiteness is considered white
+    thresh_whiteness = 100;
+    
+    %anything below thresh_blackness is considered black
+    thresh_blackness = 20;
+    c_old = [0 0];
+    
+    %extract roughly the maze from the sample image (based on whiteness)
+    if size(sample_image,3)>1
+        maze_image = rgb2graysample_image; %convert to grayscale
+    else
+        maze_image = sample_image;
+    end
+    maze_image = medfilt2(maze_image,[20,20]); %blur
+    maze_image(maze_image<thresh_whiteness) = 0; %apply threshold
+    maze_image = logical(maze_image); %binarize
+    for frame_idx = 1:num_frames
+        % Update original image CData
+        frame = video(:,:,frame_idx);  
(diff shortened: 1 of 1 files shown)


behavior/get_spatial_trace_from_behavior.m
```diff
@@ -0,0 +1,62 @@
...
+    %anything above thresh_whiteness is considered white
+    thresh_whiteness = 100;
+    
+    %anything below thresh_blackness is considered black
+    thresh_blackness = 20;
+    c_old = [0 0];
+    
+    %extract roughly the maze from the sample image (based on whiteness)
+    if size(sample_image,3)>1
+        maze_image = rgb2graysample_image; %convert to grayscale
```
kimtonyhyun: Should be `rgb2gray(sample_image)`

behavior/get_spatial_trace_from_behavior.m
```diff
@@ -0,0 +1,62 @@
...
+    else
+        maze_image = sample_image;
+    end
+    maze_image = medfilt2(maze_image,[20,20]); %blur
```
kimtonyhyun: Just a comment: Median filter is not really a "blur", or is it?

behavior/get_spatial_trace_from_behavior.m
```diff
@@ -0,0 +1,62 @@
...
+    maze_image(maze_image<thresh_whiteness) = 0; %apply threshold
+    maze_image = logical(maze_image); %binarize
```
kimtonyhyun: `maze_image = maze_image > thresh_whiteness` should do the same as the above two lines

kimtonyhyun:
Looks good to me overall. Is there a short test behavioral movie that I can apply this to?

Also, in the future, there may be some variation in the overall brightness of the behavioral movies (Herrin 224 setup will get a new camera soon, for instance). Hence, we may need to worry about adaptive threshold levels for the current behavioral cohort.

kimtonyhyun:
Also, FYI, we use <a href="https://handbrake.fr/">Handbrake</a> for encoding the behavioral movies. You should be able to use that to extract a short behavioral movie snippet, for instance.

behavior/get_spatial_trace_from_behavior.m
inanhkn: I used it to get rid of details in the scene, which I still see as a kind of blurring. I changed the comment tho.

inanhkn:
made the recommended changes

inanhkn:
I'll also write a small script that reads from a small behavior video and outputs the video overlaid with the centroids

behavior/get_spatial_trace_from_behavior.m
```diff
@@ -0,0 +1,62 @@
...
+    
+    %extract roughly the maze from the sample image (based on whiteness)
+    if size(sample_image,3)>1
+        maze_image = rgb2gray(sample_image); %convert to grayscale
+    else
+        maze_image = sample_image;
+    end
+    maze_image = medfilt2(maze_image,[20,20]); %get rid of details
+    maze_image = maze_image > thresh_whiteness; %apply threshold
+    maze_image = logical(maze_image); %binarize
```
kimtonyhyun: `maze_image` is already a logical after `maze_image > thresh_whiteness`; so, this line is redundant

PR #9, opened 2015-01-09
This file concatenates space binned tif files into an hdf5.

jmaxey:
@kimtonyhyun

### DIFF
--- a/concatenateHDF5.m
+++ b/concatenateHDF5.m
@@ -1,0 +1,50 @@
+function [hdf5Name] = concatenateHDF5(tifDir)
+% Concatenates all tif files in the specified directory into an
+% hdf5 file. A file structure similar to the one below should be used. The
+% stored files are all space binned by the factor set in 'downsmpFactor'.
+%
+% Example file structure: "G:\cohort9-herrin224\mouse5\day10_allo-south"
+% Example use: "concatenateHDF5(tifDir)" where 'tifDir' points to the
+%   folder containing the tif files
+%
+% 2015-01-06 Jessica Maxey
+
+cd(tifDir);
+list = ls('*.tif');
+firstName = strtrim(list(1,:));
+[rows,cols] = size(imread(firstName));
+numFiles = size(list,1);
+total_frames = count_frames_in_tif(tifDir);
+
+totalFrames = 1;
+[drive,buffer] = strtok(tifDir,'\');
+[cohort,buffer] = strtok(buffer,'\');
+[mouse,buffer] = strtok(buffer,'\');
+[day,buffer] = strtok(buffer,'\');
+
+hdf5Name = [cohort,'_',mouse,'_',day,'.h5'];
+dataset = ['/',mouse];
+
+downsmpFactor = 0.5;
+chunkSize = [rows*downsmpFactor cols*downsmpFactor 1];
+h5create(hdf5Name,dataset,[Inf Inf Inf],'ChunkSize',chunkSize);
+for i=1:size(list,1)
+    tifName = strtrim(list(i,:));
+    tifInfo = imfinfo(tifName);
+    tifFile = Tiff(tifName,'r');
+    numFrames = length(tifInfo);
+    for j=1:numFrames
+        tifFile.setDirectory(j);
+        imageStack(:,:,j) = imresize(tifFile.read(),downsmpFactor,'bilinear');
+    end
+    [dRows,dCols] = size(imageStack(:,:,1));
+
+    h5write(hdf5Name,dataset,imageStack,[1,1,totalFrames],[dRows,dCols,numFrames]);
+   
+    fprintf('File %d of %d stored\n',i,size(list,1));
+    totalFrames = totalFrames+numFrames;
+
+    clear imageStack tifName tifInfo numFrames
+end
+
+h5disp(hdf5Name);


concatenateHDF5.m
```diff
@@ -0,0 +1,50 @@
...
+% hdf5 file. A file structure similar to the one below should be used. The
+% stored files are all space binned by the factor set in 'downsmpFactor'.
+%
+% Example file structure: "G:\cohort9-herrin224\mouse5\day10_allo-south"
+% Example use: "concatenateHDF5(tifDir)" where 'tifDir' points to the
+%   folder containing the tif files
+%
+% 2015-01-06 Jessica Maxey
+
+cd(tifDir);
```
kimtonyhyun: Will this actually change the active directory in Matlab? If so, why not use `dir(fullfile(tifDir,'*.tif'))` which would list the tif files of the specified directory without changing the path?

concatenateHDF5.m
```diff
@@ -0,0 +1,50 @@
...
+[drive,buffer] = strtok(tifDir,'\');
+[cohort,buffer] = strtok(buffer,'\');
+[mouse,buffer] = strtok(buffer,'\');
+[day,buffer] = strtok(buffer,'\');
+
+hdf5Name = [cohort,'_',mouse,'_',day,'.h5'];
+dataset = ['/',mouse];
+
+downsmpFactor = 0.5;
+chunkSize = [rows*downsmpFactor cols*downsmpFactor 1];
```
kimtonyhyun: Does this implicitly assume `rows` and `cols` are divisible by 2?

kimtonyhyun:
Can you also provide a function that loads a hdf5 file into a movie array in memory (and, obviously, works with the hdf5 files generated by `concatenateHDF5`)?

concatenateHDF5.m
```diff
@@ -0,0 +1,50 @@
...
+[rows,cols] = size(imread(firstName));
+numFiles = size(list,1);
+total_frames = count_frames_in_tif(tifDir);
+
+totalFrames = 1;
+[drive,buffer] = strtok(tifDir,'\');
+[cohort,buffer] = strtok(buffer,'\');
+[mouse,buffer] = strtok(buffer,'\');
+[day,buffer] = strtok(buffer,'\');
+
```
kimtonyhyun: Just a comment: If you allow the user to specify the filename (as a function parameter) rather than constructing it from the path, you'd relax the requirement on the path structure.

kimtonyhyun:
Were you able to find out what caused the Matlab errors?

kimtonyhyun:
It also occurred to me just now that, if we're going to write our own Matlab wrapper to TurboReg, we don't need to have all frames in a single giant file (TIF or HDF5) for the sake of motion correction. The image registration happens on a frame-by-frame basis, and we should be able to perform motion correction even if the data is chunked into trials.

jmaxey:
So for each trial, you would load the last frame from the previous trial to start the new registration?

I'm planning to be in Herrin late this afternoon and make the corrections that you suggested. I also added code that duplicates frames when one is missing (after making the pull request). I found that it was much easier to do as the hdf5 is being written rather than in a separate function afterwards.

I think the errors I was finding were due to some kind of memory constraint/issue. Since downsampling the images prior to saving, there have been no issues.

----- Original Message -----

kimtonyhyun:
The current motion correction (Mosaic wrapper for TurboReg) registers all frames to a common reference frame, which by default is the first frame of the movie.

We could register to the immediately previous frame (after it has been registered), but I'm not sure if that will make a whole lot of difference. Intuitively, registering the entire movie to a common frame seems preferable to me as it could potentially avoid any "drift" effects over the course of the movie (if we register successive frames together).

kimtonyhyun:
Also, we have backup HDs loaded with data (for Mouse 5 and 7, at least) that you can take to CNC to work on your fancy Mac. I'll put it in front of the PC. ;)

jmaxey:
That makes sense (although we should register to the first frame that does not contain dark regions from the LED being off). Given that we can register the movies without concatenation, is there still a need to perform this step? Or would it make more sense to downsample and replace frames within each smaller tif file. I assume that it would be easy to alter the pca script to read in multiple sequential tifs rather than a single concatenated one. 

----- Original Message -----

kimtonyhyun:
It seems that you're fairly close to finishing the HDF5 concatenation, so I'd vote to complete it.

Our previous workflow (involving motion correction in Mosaic) was based on a single large movie file, so that by finishing this concatenation routine, we should be able to proceed with our old workflow. This would allow us to take a preliminary look at the new datasets without having to wait for our implementation of the TurboReg wrapper in Matlab.

We can continue to keep alive the idea of storing movies into trial segments. I'm not sure if there are going to be any major annoyances of working this way. On the other hand, we know that the single big movie file more-or-less worked, so it's good to have available as a backup.

kimtonyhyun:
Ack! I just checked and the official Inscopix distribution of Mosaic (1.0.5b) doesn't actually let us load HDF5 files...

kimtonyhyun:
@forea Did you ever get the in-lab distribution of Mosaic working on your computer?

kimtonyhyun:
Is it possible to not _have_ to downsample on the HDF5 concatenation? It seems to me that there could be cases where we want to work with the 720 x 540, rather than the miniscule 360 x 270. With 100 trials, it seems that the 720 x 540 movies will be ~30 GB each, which seems like a reasonable size given our compute machines.

jmaxey:
I can try but every time I attempted to do this before, matlab would crash due to unknown errors (that I assumed were memory related). I was not able to find the exact source of the error but can keep looking for it.

----- Original Message -----

kimtonyhyun:
You should also give it a try on your computer as well. We're finding some weird HD error issues with the big PC at Herrin (prevents us from copying files), so there's a (slight) chance that your error could be related.

concatenateHDF5.m
```diff
@@ -0,0 +1,50 @@
...
+chunkSize = [rows*downsmpFactor cols*downsmpFactor 1];
+h5create(hdf5Name,dataset,[Inf Inf Inf],'ChunkSize',chunkSize);
+for i=1:size(list,1)
+    tifName = strtrim(list(i,:));
+    tifInfo = imfinfo(tifName);
+    tifFile = Tiff(tifName,'r');
+    numFrames = length(tifInfo);
+    for j=1:numFrames
+        tifFile.setDirectory(j);
+        imageStack(:,:,j) = imresize(tifFile.read(),downsmpFactor,'bilinear');
```
kimtonyhyun: You should pre-allocate `imageStack`. I think this could be related to the crash.

PR #10, opened 2015-01-17
Kimth/io

kimtonyhyun:
@forea

The function `compare_frames_in_xml_tif(...)` helps to make sure that RAW file decompression didn't glitch, by checking the frame counts of each XML and TIFF file pair.

If a mismatch is detected, an error is issued. In all cases so far, re-decompressing the RAW file has taken care of such mismatches.

### DIFF
--- a/io/compare_frames_in_xml_tif.m
+++ b/io/compare_frames_in_xml_tif.m
@@ -16,7 +16,7 @@ for i = 1:num_files
     [~, name, ~] = fileparts(xml_filename);
     tif = dir(fullfile(path_to_xml, strcat(name, '.tif')));
     if (isempty(tif))
-        fprintf('  Error! "%s" is missing its RAW or TIF counterpart!\n',...
+        fprintf('  Error! "%s" is missing its TIF counterpart!\n',...
             xml_filename);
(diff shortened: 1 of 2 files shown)


io/compare_frames_in_xml_tif.m
```diff
@@ -0,0 +1,42 @@
...
+    tif = dir(fullfile(path_to_xml, strcat(name, '.tif')));
+    if (isempty(tif))
+        fprintf('  Error! "%s" is missing its TIF counterpart!\n',...
+            xml_filename);
+        mismatch_detected = true;
+        break;
+    else
+        tif_filename = fullfile(path_to_xml, tif.name);
+        tif_info = imfinfo(tif_filename);
+        num_tif_frames = length(tif_info);
```
forea: Just trying to understand the code here -- Does imfinfo(tif_filename) return info for each frame? And that's why num_tif_frames = length(tif_info)?

kimtonyhyun: Yep. In the past I wanted to find a quicker way of finding the length of a TIFF stack, but this was all I found.

PR #11, opened 2015-01-17
Kimth/bigtiff

kimtonyhyun:
@forea @jmaxey @inanhkn 

So that we can start taking a look at Cohort 9, I've modified the BigTiff concatenation routine (as found in Dropbox/herrin/scripts) to be compatible with the new recording format -- where we only record during trials. This generates a single TIFF file that is compatible with Mosaic so that we can proceed with motion correction through Mosaic as before.

The main routine for concatenating into a single BigTiff is `concatenate_bigtiff`. It reads in the output generated by the plus maze (the text file with trial info), and a path to a directory containing XML/TIF pairs, as well as a `trim` parameter. The function will generate a single large Tiff file keeping only valid trial frames. Furthermore, it will trim a few frames from the beginning and end of each trial as specified by `trim`. This is necessary because the first few frames of every trial is dark (the Miniscope starts recording before the LED has fully turned on).

The companion function to `concatenate_bigtiff` is `compress_frame_indices`, which will regenerate frame indices to be used with the concatenated movie. Note that `compress_frame_indices` needs to be invoked with the same `trim` parameter as in the movie concatenation.

I'll send out another note once I've generated some concatenated files from Cohort 9. Right now, the big PC at Herrin is busy copying files to backup hard drives.


kimtonyhyun:
I was able to concatenate data from Cohort 9 (Mouse 2, Day 9) as a single BigTiff file. Running through the same pipeline as before...

jmaxey:
These look good. I will incorporate is_trial_frame() into the hdf5 process so that false trials are excluded from the final file. Does the script output a list of new trial frames since they are altered by dropping the false trials? If so, I did not see where this happens.

kimtonyhyun:
Right now, `concatenate_bigtiff` does not generate the list of new trial frames. Instead, that is done in a separate script `compress_frame_indices`, which is to be invoked with the same `trim` parameters as the concatenation function. I agree that this separation is nonoptimal. Ideally, we would store the new indices along with the concatenated movie (should be possible with HDF5 storage?).

jmaxey:
I will incorporate it into the HDF5 storage.

PR #13, opened 2015-01-28
Kimth/classify ics

kimtonyhyun:
Port of the manual classification function from Dropbox/herrin/scripts. The top-level function is `classify_ics`. Unlike the version in Dropbox, the frame indexing is now consistent with Cohort 9 data. I've attached a readme in the `classification` folder that briefly describes the role of each function.

@forea We are seeing trial phase-sensitive cells in Cohort 9, as before (Mouse 7, Day 7 shown):
c9m7_d07


kimtonyhyun:
Added support for movies stored as HDF5, where the loading is _way_ faster.

kimtonyhyun:
@inanhkn Ping; can you review?

kimtonyhyun:
Actually, the various `view_...` functions used exclusively by `classify_ics` may be better organized as internal / nested functions of `classify_ics`. I'll ping again after making the change.

kimtonyhyun:
@inanhkn I moved the various `view_...` functions into a `support` subdirectory. I also added brief comments to teach function. Please review.

By the way, using `class` as a variable name is not good (shadows the Matlab built-in function). Let's factor that out in the future.

inanhkn:
It looks good to me. Thanks for having the movie array itself as an input in run_pca, I've wasted a total of 2 hours or so because of the former structure so far :)

inanhkn:
I'm not sure about the folder structure, I would never be sure when to put something into a folder named utils for example:) But I'm sure you have thought it out better than I would.

kimtonyhyun:
Regarding loading movies in `run_pca` we are pretty close to transitioning to HDF5 storage now. We've found that you can load the typical Cohort 9 movie in about 10 seconds -- so less time wasted.

PR #14, opened 2015-01-28
Basic view movie script

kimtonyhyun:
Basic script to view a movie stack in Matlab. We should further develop this so that we can at least thumb through the frames using a slider.

### DIFF
--- a/view/view_movie.m
+++ b/view/view_movie.m
@@ -1,0 +1,15 @@
+function view_movie(M)
+
+num_frames = size(M,3);
+
+h = imagesc(M(:,:,1));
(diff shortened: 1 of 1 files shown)


view/view_movie.m
```diff
@@ -0,0 +1,15 @@
+function view_movie(M)
```
forea: for sake of completeness, should we explicitly state expected input M?

kimtonyhyun: Ok ok...

kimtonyhyun:
Addressed comment. Merging.

PR #15, opened 2015-01-28
Kimth/hdf5

kimtonyhyun:
@jmaxey Generic saving and loading of movies (three-dimensional matrices `[height width num_frames]`) via HDF5 files. Uses a default dataset name of `/movie` (we can adopt a different standard, feel free to suggest).

For completeness, I also added a script that saves a movie in memory to BigTiff format that is compatible with Mosaic. (The script for loading a TIFF file into memory is already in master.)

### DIFF
--- a/io/load_movie_from_hdf5.m
+++ b/io/load_movie_from_hdf5.m
@@ -1,0 +1,7 @@
+function M = load_movie_from_hdf5(source)
+% Load a movie matrix [height x width x num_frames] from a HDF5 file.
+%   Uses a default dataset name of "/movie"
+%
+% 2015 01 28 Tony Hyun Kim
+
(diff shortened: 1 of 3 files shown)


kimtonyhyun:
I'd also like to provide a default `dataset_name` when none is provided (save typing during use). Should we go with `/Data/Images` for now?

kimtonyhyun:
I added the dataset name as an externally accessible parameter.

Regarding the size of the dataset: My intended use case for the HDF5 file is to store relatively simple movies (hence the name, load and save _movie_). I don't expect to add arbitrarily sized objects to the same dataset containing the movie. Furthermore, if needed, we could add additional parameters to other datasets of the same HDF5 file.

PR #16, opened 2015-01-30
Concatenate all frames from a trial into an hdf5

jmaxey:
@kimtonyhyun

### DIFF
--- a/io/concatenateHDF5.m
+++ b/io/concatenateHDF5.m
@@ -1,0 +1,157 @@
+function concatenateHDF5(tifDir,outputDir,hdf5Name,downsmpFactor,plusmazeName,trim)
+
+% Concatenates all tif files in the specified directory into an
+% hdf5 file. The stored files are all downsampled by the factor set in 
+% 'downsmpFactor'. The number of frames to be dropped from the beginning 
+% and end of each trial is set in 'trim'. Bad trials are removed (using the
+% plusmaze text file) and if frames were dropped during a good trial, the 
+% previous frame is used to fill the gap. A new text file is produced
+% indicating the start and stop frames for each trial (as well as the gate
+% up/down frames, the direcitons of the trial and time to complete each
+% trial).
+%
+% Example arguments:
+% tifDir = '/Volumes/COHORT9/cohort9-herrin224/mouse5/day20_ego-right';
+% outputDir = '/Users/jmaxey/Documents/MATLAB/PreFrontal';
+% hdf5Name = 'test.hdf5';
+% downsmpFactor = 0.5;
+% plusmazeName = 'mouse5_day20_ego-right.txt';
+% trim = [10,5];
+%
+% 2015-01-27 Jessica Maxey
+
+outputFileName = fullfile(outputDir,[plusmazeName(1:end-4),'_new.txt']);
+outputFile = fopen(outputFileName,'wt+');
+
+[frame_indices,location_info,time] = getTrialIndicies(fullfile(tifDir,plusmazeName));
+startFrames = frame_indices(:,1);
+
+list = dir(fullfile(tifDir,'*.tif'));
+firstName = fullfile(tifDir,list(1).name);
+[rows,cols] = size(imread(firstName));
+numFiles = length(list);
+
+totalFrames = 1;
+
+downsmpRows = floor(rows*downsmpFactor);
+downsmpCols = floor(cols*downsmpFactor);
+chunkSize = [downsmpRows downsmpCols 1];
+dataset = '/Data/Images';
+h5create(fullfile(outputDir,hdf5Name),dataset,[Inf Inf Inf],'ChunkSize',chunkSize,'Datatype','uint16');
+badTrial = 0;
+skipCount = 0;
+testFrames = 1;
+trialCount = 0;
+
+for i=1:length(list)
+    if(i ~= 1)
+        %%% Find if the start frame of the next recording matches one of
+        %%% the start frame indicies
+        idx = find(testFrames == startFrames);
+        if(isempty(idx))
+            %%% Bad trial
+            badTrial = 1;
+            tif_info = imfinfo(fullfile(tifDir,list(i).name));
+            numBadFrames = length(tif_info);
+            testFrames = testFrames + numBadFrames;
+        else
+            %%% Good trial
+            badTrial = 0;
+        end
+    end
+
+    if(badTrial == 0)
+        %%% Add the good trial to the hdf5 file
+        trialCount = trialCount+1;
+        tifName = fullfile(tifDir,list(i).name);
+        tifInfo = imfinfo(tifName);
+        tifFile = Tiff(tifName,'r');
+        oriFrames = length(tifInfo);
+        
+        xmlName = [tifName(1:end-3),'xml'];
+        xmlData = parse_miniscope_xml(xmlName);
+        
+        %%% Determine if frames were dropped by the miniscope during the
+        %%% trial
+        if(str2num(xmlData.dropped_count) ~= 0)
+            droppedFrames = str2num(xmlData.dropped);
+        else
+            numFrames = oriFrames;
+            droppedFrames = 0;
+        end
+        
+        %%% Read in the images, trim the frames at the beginning and end 
+        %%% of the trial (set by the 'trim' argument), and downsample by the 
+        %%% 'downsmpFactor' argument
+        imageStack = uint16(zeros(floor(rows*downsmpFactor),floor(cols*downsmpFactor),oriFrames-trim(1)-trim(2)));
+        if(downsmpFactor == 1)
+            for j=1:oriFrames-trim(1)-trim(2)
+                tifFile.setDirectory(j+trim(1));
+                imageStack(:,:,j) = uint16(tifFile.read());
+            end
+        else
+            for j=1:oriFrames-trim(1)-trim(2)
+                tifFile.setDirectory(j+trim(1));
+                imageStack(:,:,j) = uint16(imresize(tifFile.read(),downsmpFactor,'bilinear'));
+            end
+        end
+        
+        %%% Replace any dropped frames with the frame immediately
+        %%% preceeding it
+        if(droppedFrames ~=0)
+            for j=1:length(droppedFrames)
+                disp('DroppedFrame');
+                droppedFrame = droppedFrames(j);
+                if((droppedFrame > trim(1)) && (droppedFrame < oriFrames-trim(2)))
+                    droppedFrame = droppedFrame+trim(1);
+                    frontStack = imageStack(:,:,1:droppedFrame-1);
+                    frontStack = cat(3,frontStack,imageStack(:,:,droppedFrame-1));
+                    backStack = imageStack(:,:,droppedFrame:end);
+                    newImageStack = cat(3,frontStack,backStack);
+                    clear imageStack
+                    imageStack = newImageStack;
+                end
+            end
+        end
+        
+        numFrames = size(imageStack,3);
+        [dRows,dCols] = size(imageStack(:,:,1));
+        
+        h5write(fullfile(outputDir,hdf5Name),dataset,imageStack,[1,1,totalFrames],[dRows,dCols,numFrames]);
+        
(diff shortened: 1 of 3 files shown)


io/concatenateHDF5.m
```diff
@@ -0,0 +1,157 @@
...
+% tifDir = '/Volumes/COHORT9/cohort9-herrin224/mouse5/day20_ego-right';
+% outputDir = '/Users/jmaxey/Documents/MATLAB/PreFrontal';
+% hdf5Name = 'test.hdf5';
+% downsmpFactor = 0.5;
+% plusmazeName = 'mouse5_day20_ego-right.txt';
+% trim = [10,5];
+%
+% 2015-01-27 Jessica Maxey
+
+outputFileName = fullfile(outputDir,[plusmazeName(1:end-4),'_new.txt']);
```
kimtonyhyun: FYI, the built-in function `fileparts` will separate the file name from the extension. Useful for cases like TIF files which may have `.tiff` or `.tif` as the extension.

io/concatenateHDF5.m
```diff
@@ -0,0 +1,157 @@
...
+outputFile = fopen(outputFileName,'wt+');
+
+[frame_indices,location_info,time] = getTrialIndicies(fullfile(tifDir,plusmazeName));
```
kimtonyhyun: The function `behavior/get_trial_frame_indices` already performs this task.

kimtonyhyun:
Instead of saving a new text file with the recomputed frame indices, why don't we just embed it in the HDF5 file itself (just the recomputed frame indices, not all the info like begin / end / time of trial).

io/concatenateHDF5.m
```diff
@@ -0,0 +1,157 @@
...
+        
+        numFrames = size(imageStack,3);
+        [dRows,dCols] = size(imageStack(:,:,1));
+        
+        h5write(fullfile(outputDir,hdf5Name),dataset,imageStack,[1,1,totalFrames],[dRows,dCols,numFrames]);
+        
+        %%%%%%%%%%%%%% NEED TO FIX %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+        %%% Currently does not take into account added frames if some were
+        %%% originally dropped - only tested on cases without dropped
+        %%% frames
```
kimtonyhyun: Let's operate with the assumption that the text file from the plus maze has the absolutely correct frame counts. (In cases where the text file and the Miniscope differ, we can manually edit the text file to correct the discrepancy.)

Also, `io/compress_frame_indices.m` already takes care of recomputing trial frame indices, given the original frame indices and the `trim` parameter.

io/concatenateHDF5.m
```diff
@@ -0,0 +1,157 @@
...
+        testFrames = testFrames+oriFrames;
+        
+        clear imageStack tifName tifInfo tifFile droppedFrames newImageStack
+    end
+end
+
+%%% Remove the extra frame count needed to index the hdf5 file
+totalFrames = totalFrames-1;
+
+h5create(fullfile(outputDir,hdf5Name),'/Parameters/Number of Frames',1);
```
kimtonyhyun: Can we use shorter names without spaces, e.g. `/Params/numFrames'?

jmaxey:
If we are getting rid of beginning/end/time, do you want me to store a new variable showing if the mouse got the trial correct or not? Also, if we are going to try to normalize the trials later on, having the time it took to complete the trial could be useful. Either that or we should store the frame rate of the miniscope so that we can find the trial time using the frame indicies (that would probably be better in the long run). 

----- Original Message -----

kimtonyhyun:
Some minor comments. I'm checking out the HDF5 code on the Big PC at Herrin -- where I'll try to concatenate Mouse 7 Day 8 (60k frames). If the HDD issue becomes a problem, then I'll try out on Hakan's computer.

io/concatenateHDF5.m
jmaxey: So we should assume that the original text file will have correct frames while the movie itself might have dropped frames so that after the frames are added to the movie, the current code will be correct (just want to be sure I understand).

----- Original Message -----

kimtonyhyun:
The HDF5 file can also store beginning/end/time info for completeness. However, I'd suggest that we store it in under a different dataset (e.g. `/Params/trialInfo`). I suspect that we'll need to read in the trial frame indices frequently, so it'd be good to have it already in a numeric array format (rather than string text that needs to be parsed all the time).

io/concatenateHDF5.m
kimtonyhyun: Yes, we are in agreement.

I think this assumption handles most cases. When the assumption is false, we'll manually fix up the text file.

io/concatenateHDF5.m
```diff
@@ -0,0 +1,157 @@
...
+            tif_info = imfinfo(fullfile(tifDir,list(i).name));
+            numBadFrames = length(tif_info);
+            testFrames = testFrames + numBadFrames;
+        else
+            %%% Good trial
+            badTrial = 0;
+        end
+    end
+
+    if(badTrial == 0)
```
kimtonyhyun: Minor suggestion: another way to write this is `~badTrial`

io/concatenateHDF5.m
```diff
@@ -0,0 +1,157 @@
...
+            droppedFrames = str2num(xmlData.dropped);
+        else
+            numFrames = oriFrames;
+            droppedFrames = 0;
+        end
+        
+        %%% Read in the images, trim the frames at the beginning and end 
+        %%% of the trial (set by the 'trim' argument), and downsample by the 
+        %%% 'downsmpFactor' argument
+        imageStack = uint16(zeros(floor(rows*downsmpFactor),floor(cols*downsmpFactor),oriFrames-trim(1)-trim(2)));
```
kimtonyhyun: Should write: `zeros(downsmpRows, downsmpCols, oriFrames-trim(1)-trim(2), 'uint16')`. This avoids allocating a `double` array then converting to `uint16`. Also, you previously defined the floored downsampled dimensions.

io/concatenateHDF5.m
```diff
@@ -0,0 +1,157 @@
...
+            end
+        else
+            for j=1:oriFrames-trim(1)-trim(2)
+                tifFile.setDirectory(j+trim(1));
+                imageStack(:,:,j) = uint16(imresize(tifFile.read(),downsmpFactor,'bilinear'));
+            end
+        end
+        
+        %%% Replace any dropped frames with the frame immediately
+        %%% preceeding it
```
kimtonyhyun: Slicing the movie array to insert dropped frames this way is a bit inefficient (better to write it as `imageStack` is originally filled in). However, there aren't that many dropped frames, so this probably doesn't matter in practice.

jmaxey: I can try to think of another way to do this, however, the main problem that I ran into is that the dropped frames are normally sequential. If you just use indexing within the entire stack, you end up pulling frames after the dropped frame instead of the one just before it.

----- Original Message -----

kimtonyhyun: I wouldn't worry about it for now. Maybe just write a `TODO: ...` in the comments.

kimtonyhyun:
By the way, good idea to save the frame rate into `/Params` as well.

io/concatenateHDF5.m
```diff
@@ -0,0 +1,157 @@
...
+% outputDir = '/Users/jmaxey/Documents/MATLAB/PreFrontal';
+% hdf5Name = 'test.hdf5';
+% downsmpFactor = 0.5;
+% plusmazeName = 'mouse5_day20_ego-right.txt';
+% trim = [10,5];
+%
+% 2015-01-27 Jessica Maxey
+
+outputFileName = fullfile(outputDir,[plusmazeName(1:end-4),'_new.txt']);
+outputFile = fopen(outputFileName,'wt+');
```
kimtonyhyun: Need to close the file at the end of script (`fclose(outputFile)`)

kimtonyhyun:
Checked it out with one of the days (Mouse 2, Day 9). Worked nicely -- and compatible with `io/load_movie_from_hdf5` ;)

Note: when you save `dRows`, etc. into `Params` as `uint16`, Matlab generates a warning: `Conversion from double to datatype H5T_STD_U16LE may colamp some values`. You can probably suppress the warning by explicitly casting the variable into `uint16` when calling `h5write`.

Looks good otherwise. After you address the minor comments, let's merge. I'll send the PR for TurboReg afterwards, and we'll operate completely through HDF5 from now on.

io/concatenateHDF5.m
```diff
@@ -0,0 +1,157 @@
...
+h5create(fullfile(outputDir,hdf5Name),'/Parameters/Number of Frames',1);
+h5write(fullfile(outputDir,hdf5Name),'/Parameters/Number of Frames',totalFrames);
+h5create(fullfile(outputDir,hdf5Name),'/Parameters/Number of Rows',1,'Datatype','uint16');
+h5write(fullfile(outputDir,hdf5Name),'/Parameters/Number of Rows',dRows);
+h5create(fullfile(outputDir,hdf5Name),'/Parameters/Number of Columns',1,'Datatype','uint16');
+h5write(fullfile(outputDir,hdf5Name),'/Parameters/Number of Columns',dCols);
+h5create(fullfile(outputDir,hdf5Name),'/Parameters/Trim Values',[1 2],'Datatype','uint16');
+h5write(fullfile(outputDir,hdf5Name),'/Parameters/Trim Values',trim);
+h5create(fullfile(outputDir,hdf5Name),'/Parameters/Downsample Factor',1,'Datatype','uint16');
+h5write(fullfile(outputDir,hdf5Name),'/Parameters/Downsample Factor',downsmpFactor);
```
kimtonyhyun: `downsmpFactor` is definitely not an integer type. Should be `double`

jmaxey: Good catch! I will change it.

io/concatenateHDF5.m
jmaxey: This file is slightly different. The existing one only pulled the start and end frames of the trial.

kimtonyhyun: Do you mean that it only pulls the frame indices and not the start/end/time information? The code `behavior/get_trial_frame_indices.m` returns all frame indices, including opening and closing of gates.

jmaxey: Ah. I thought it was just start and end. As I am using all of the fields though, is it ok for me to keep the other function? 

----- Original Message -----

kimtonyhyun: In that case, it should be called `parse_plusmaze` or something to the effect (since it returns more than the indices), and we can deprecate the existing functions.

io/concatenateHDF5.m
jmaxey: At this point, I do not think the matlab function allows non numeric data to be stored in an hdf5. Please let me know if you have found something different? Would you want to create a numeric mapping for the locations?
1 = north
2 = south
3 = east
4 = west
or something like that? If not, I could include a parameter saying if the trial was correct and not include the actual locations (although it seem that they will be needed later in the analysis).

jmaxey: Once this is clarified, I will update the PR.

kimtonyhyun: In this case, let's just leave out the start/goal/end/time information from the HDF5. The numeric mapping could be confusing. So, for the time being, we'd have to go back to the plus maze text file to get this information.

It seems that one can write string datasets, but not using `h5write`. See "Limitations" on the Matlab [page](http://www.mathworks.com/help/matlab/ref/h5d.create.html). This could be added in a later PR.

jmaxey: Did you see the updated PR? I'm not sure that I did it correctly.

kimtonyhyun:
Once you request a PR from a branch, any subsequent commits you make to the branch are automatically appended to the PR (i.e. you don't need to submit a new request). Looks good to me, let me know when you're done editing and I'll merge.

jmaxey:
I'm done with the edits. I left the trimming in the main script instead of calling the outside function since there are parameters I use for finding the final frame values that change within every iteration.

I also had already implemented the conversion of the locations to numbers when I committed the new version. I don't think leaving these in for now will hurt anything and I can change them in a future PR when I figure out how to store the actual strings. Let me know if this isn't ok.

----- Original Message -----

PR #17, opened 2015-02-04
Kimth/memory efficient pca

kimtonyhyun:
@inanhkn Minor tweak to PCA routine. (And a script to temporally bin movies.)

The motivation was to be able to run PCA on non-downsampled (neither space nor time) movies of Cohort 9 Mouse 7. For reference, the movie turned out to be 40 GB.

The PCA computation process has been modified so that we do not duplicate the movie object in memory during computation. This is crucial when the movie matrix is 40 GB, and we have 64 GB workstations.

Details:
- In the previous structure, we ended up making two copies of the movie object in memory during the PCA computation process, because the movie was instantiated in the scope of `run_pca.m`, and then modified (reshaped and column-mean centered) in the inner function `compute_pca.m`.
- In the new code, the outer function `run_pca.m` instantiates the movie matrix, and does an _in place_ reshape and column-mean centering. Note that the inner function `compute_pca.m` does not ever write to the matrix `M`, which prevents the movie duplication.
- For more details, see [Matlab copy on write](http://blogs.mathworks.com/loren/2006/05/10/memory-management-for-functions-and-variables/)


kimtonyhyun:
I'll give `bsxfun` a try and ping this thread after.

kimtonyhyun:
Now uses `bsxfun` for the per-frame normalization. However, uses `compute_mean_fluorescence` which has a `for` loop anyway... Please re-review.

kimtonyhyun:
@inanhkn Removed mean-subtraction from PCA routine

PR #18, opened 2015-02-04
3 movie normalization routines to be applied sequentially

inanhkn:
@kimtonyhyun 

Added 3 normalization routines. 
1) fit_mean_fluorescence.m: inverts the exponential trend
    -uses compute_mean_fluorescence.m to calculate the mean fluorescence

2) norm_by_disk_filter.m : normalizes frames by the disk filter output

3) norm_by_background_mean.m: normalizes by the mean background fluorescence(would probably be ok to simply normalize by the total mean fluorescence)

### DIFF
--- a/image_op/norm_by_background_mean.m
+++ b/image_op/norm_by_background_mean.m
@@ -1,0 +1,22 @@
+function M_norm = norm_by_background_mean(movie)
+%Normalizes every frame in the movie by the mean pixel value of the
+%background to mitigate the temporal variation in the brightness
+%
+%   movie: movie matrix , [h x w x num_frames]
+%
+% Hakan Inan, 15-Jan-4
+%
+
+M_norm = zeros(size(movie),'single');
+[height,width,numFrames] = size(movie);
+
+for k = 1:numFrames
+    if (mod(k,1000)==0)
+        fprintf('  Frames %d of %d done\n', k, numFrames);
+    end
+    Frame = reshape(movie(:,:,k),height*width,1);
+    thresh = quantile(Frame,0.7);
+    Z = mean(Frame(Frame<thresh));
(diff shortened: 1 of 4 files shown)


inanhkn:
Fyi, removing the exponential trend becomes obsolete after the addition of norm_by_background_mean to the pipeline

image_op/norm_by_background_mean.m
```diff
@@ -0,0 +1,22 @@
...
+
+M_norm = zeros(size(movie),'single');
+[height,width,numFrames] = size(movie);
+
+for k = 1:numFrames
+    if (mod(k,1000)==0)
+        fprintf('  Frames %d of %d done\n', k, numFrames);
+    end
+    Frame = reshape(movie(:,:,k),height*width,1);
+    thresh = quantile(Frame,0.7);
```
kimtonyhyun: The quantile parameter (`0.7`) should be exposed as a function parameter (e.g. `quantile_threshold`). Also, when `quantile_threshold == 1` then you should skip the `quantile` operation and perform a naive mean without regards to histogram.

image_op/norm_by_disk_filter.m
```diff
@@ -0,0 +1,24 @@
...
+%
+%   movie: movie matrix , [h x w x num_frames]
+%
+% 2015 01 31 Tony Hyun Kim (Revised: Hakan Inan, 15-Jan-4)
+%
+
+M_norm = zeros(size(movie), 'single');
+num_frames = size(movie,3);
+% Apply spatial normalization
+hDisk = fspecial('disk', 15);
```
kimtonyhyun: The radius of the disk should be exposed as a function parameter (e.g. `disk_radius`).

inanhkn:
disk_radius and quantiles can now be passed as input arguments (optional).

PR #19, opened 2015-02-05
Inanhkn/remove autoscaling movie

inanhkn:
@kimtonyhyun @forea 

In `view_ic_over_movie_interactively.m`, I scaled the movie roughly to `[0,1]` and set `CLim` parameters for `imagesc(...)` accordingly. Since every frame is now scaled by the same constant, traces should be more interpretable.

### DIFF
--- a/classification/support/view_ic_over_movie_interactively.m
+++ b/classification/support/view_ic_over_movie_interactively.m
@@ -10,23 +10,33 @@ mad_scale = 8; % Used for coarse detection of activity in the IC trace
 active_frame_padding = padding; % Use 100 for 20 Hz movie
 time_window = 10; % Width of running window
 
+%Custom Scaling for movie display
+%-----------------------------------------------------------
+%CLim parameters for imagesc(). The movie is scaled to [0,1], so the most
+%natural choice for [clim_lower,clim_upper] = [0,1]. However, it might be
+%set to values in the neighborhood to have darker or higher-contrast image.
+clim_lower = 0;
+clim_upper = 1.4;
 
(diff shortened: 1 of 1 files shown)


kimtonyhyun:
Looks good to me, although we'll probably want to expose the `CLim` params at runtime (so the fixed scale can be adjusted on the fly).

classification/support/view_ic_over_movie_interactively.m
```diff
@@ -10,15 +10,44 @@
...
+
+threshUpBright = quantile(maxVec,quant_upper);
+threshDownBright = quantile(maxVec,quant_lower);
+
+threshUpDark = quantile(minVec,quant_upper);
+threshDownDark = quantile(minVec,quant_lower);
+
+ZHigh = mean(maxVec(maxVec>threshDownBright & maxVec<threshUpBright));
+ZLow = mean(minVec(minVec>threshDownDark & minVec<threshUpDark));
+
```
kimtonyhyun: Performing the above set of operations _for each IC_ is expensive and has considerably slowed down the load time of each plot.

inanhkn:
I think I was a bit silly to put this in `view_ic_over_movie_interactively.m ` . This can be done once in `classify_ics.m` and the scaled movie can be passed to this script. We don't need to calculate `ZHigh, ZLow` for each IC.

kimtonyhyun:
Yeah no worries, I have a fix for this,

-Tony

2015-02-08 14:28 GMT-08:00 inanhkn <notifications@github.com>:

PR #21, opened 2015-02-09
Kimth/turboreg

kimtonyhyun:
@jmaxey @inanhkn PR to enable TurboReg from Matlab.

There are a lot of files in this PR, but most of it is the C source code for TurboReg. Only the following files are relevant for us:
- `M_reg = run_turboreg(M)`: Apply TurboReg registration to movie `M`. This is essentially a port of the registration process as implemented in Mosaic.
- `turbocoreg.c`: This is the gateway MEX function into TurboReg. (MEX allows Matlab to make calls to compiled C code.) This has been significantly edited from Jerome/Biafra's implementation -- to get rid of unnecessary and distracting options into TurboReg, some minor bug fixes, and to merge the function of two MEX files into one (they used two separate C files `turboreg` and `transfturboreg`).
- `compare_registered_movie(M,M_reg)` will display two movies side-by-side with locked XY limits, so that one can compare the success of the registration process.

Note: in order to use the MEX function `turbocoreg`, you need to compile the MEX code on your local machine. The instructions to do so are in the leading comments of `turbocoreg.c`. If you need any help in compiling the MEX code, let me know.

Once you've compiled the MEX function, you can try out the TurboReg registration as follows:

1. Download a short movie snippet. I found Cohort 8 Mouse 1 to be a nice example, because he has a very strong blood vessel structure in the center of the FOV. [Link to a c8m1 snippet](https://www.dropbox.com/s/yzkz1g5yij2xu0y/c8m1d8_snippet.mat?dl=0). This `.mat` file produces a short (200 frame) movie matrix `M` when loaded in Matlab.
2. Run TurboReg on this short movie by running `M_reg = run_turboreg(M)`. The script will prompt for some parameters (e.g. which frame to use as reference). Just press enter, to fill each parameter with the defaults. The script also prompts the user to draw an ROI for the registration process (see the image below)
3. Examine the registered movie against the original by running `compare_registered_movie(M,M_reg)`. Use the zoom tool to inspect the movie in more detail.

20150208_turboreg_roi

For reference, the 200 frame movie takes about 20 seconds to register on my laptop. Currently, the registration script runs single-threaded. Down the line, we should add parallel frame processing.


kimtonyhyun:
Here is a bit more explanation of how `turbocoreg` process works:
turbocoreg

At the end of the day, image registration is a process that takes two images (reference image, and image to be registered). These inputs are denoted in green at the top of the flowchart.

Now, the registration process from Mosaic was doing a bit more than straight-up registration of these two input images. Mosaic's registration process allowed the user to specify some pre-registration image filtering of the two source images (mostly blurring), and fed the processed images to the registration routine. This is denoted by the "Some filtering" (pink box) in the flowchart.

(I experimented with these filtering operations -- including doing nothing -- and experimentally determined that they _do_ have a big effect in the success of the overall registration.)

TurboReg then determines the transformation parameters to transform the filtered image to the filtered reference image under the specified ROI. These transformation parameters are then applied to the unfiltered image ("Coreg") in order to produce the final output.

kimtonyhyun:
@jmaxey Have you had a chance to review this PR? (I'm running this routine on c9m7d08 right now -- we'll see how well it works...)

kimtonyhyun:
Didn't work super well. Trying again without rotational DOF...

kimtonyhyun:
Also FYI, Fori noted some

kimtonyhyun:
Translation only correction looks okay on spot check (whereas the uncorrected movie showed clear jitter).

kimtonyhyun:
FYI, compiling the TurboReg C binaries on Matlab 2014b required a bit of tweaking. See [this post](http://stackoverflow.com/questions/26426160/matlab-2014b-no-more-supports-user-defined-mex-configuration/26435218#26435218) for setting compiler flags.

@inanhkn On Ubuntu 14.10, I found that it was necessary to downgrade `gcc` to version 4.7 (from 4.9) in order to run `mex`, and furthermore I had to explicitly change the compiler flag to enable the modern C conventions (i.e. `-std=c99` rather than `-ansi`). See more [here](http://blogs.bu.edu/mhirsch/2013/07/matlab-r2013a-mex-on-ubuntu-13-04-64-bit/).

jmaxey:
I think I have it working now (I'm waiting for some EM stuff to finish running before testing it). After changing the -ansi to -std=c99, I also changed the imac64 sdk version to 10.9. I think the version changes with Xcode upgrades and has to be altered in the mexopts.sh file.

----- Original Message -----

jmaxey:
It works on my Mac! 

I also tested the FFT motion correction that Mark sent the link to. It does seem to help but TurboReg seems to do a much better job in one of the data.

----- Original Message -----

kimtonyhyun:
Cool -- one thing to keep in mind that TurboReg performs registration on the filtered version of the reference and image to be registered. It's not a fair comparison unless we apply the same preprocessing to the other MC methods.

PR #24, opened 2015-02-11
event extraction scripts

inanhkn:
@kimtonyhyun 

event extraction routine. Run `extract_events_2step.m` with the input whose format is explained in the script.

Note: the script is in no way optimized for anything, so expect inefficiency.


signal_op/SG_filter_trace.m
```diff
@@ -0,0 +1,6 @@
+function out = SG_filter_trace(in,N,L)
+ h = SGsmoothingfilter(N,L);
+out = conv(h,in); %apply SG filter on the raw data
+out = out(N+1:end-N); %leave out the transients of the filtering
```
kimtonyhyun: Are you trying to get the same effect as `conv(...,'same')`? Also, delete trailing newlines.

kimtonyhyun:
Note: This PR is for review only -- do not merge.

inanhkn:
The event detection script is now ready to review to be merged.
Usage:
`event_times = extract_events(trace)` This will output the event indicesin the input trace as a [# of events] x 1 array.
`event_times = extract_events(trace,1)` will also generate 2 plots like below

h

res

res2

inanhkn:
Also I removed something very redundant in `SGsmoothingfilter.m` and made `extract_candidate_events.m` so that the event magnidute threshold can be tweaked outside the function.
Now call `extract_candidate_events.m` with extract_candidate_events.m(trace,mad_scale), where mad_scale is a constant that is multiplied with the  mad of the trace to obtain the event magnitude threshold.

kimtonyhyun:
When I run this code on a good trace of c9m7d08, I get warnings like:
```
>> event_times = extract_events(trace,1);
Warning: Matrix is close to singular or badly scaled. Results may be inaccurate. RCOND =  9.073644e-32. 
> In SGsmoothingfilter at 43
  In extract_events>iterate_event_indices at 98
  In extract_events at 30 
Warning: First element of input column does not match first element of input row.
         Column wins diagonal conflict. 
> In toeplitz at 32
  In extract_events>invert_h at 149
  In extract_events>iterate_event_indices at 101
  In extract_events at 30 
Warning: Matrix is close to singular or badly scaled. Results may be inaccurate. RCOND =  9.073644e-32. 
> In SGsmoothingfilter at 43
  In extract_events>iterate_event_indices at 98
  In extract_events at 32 
Warning: First element of input column does not match first element of input row.
         Column wins diagonal conflict. 
> In toeplitz at 32
  In extract_events>invert_h at 149
  In extract_events>iterate_event_indices at 101
  In extract_events at 32 
```

signal_op/extract_events.m
```diff
@@ -0,0 +1,174 @@
+function event_times = extract_events(trace,varargin)
+% event_times = extract_events(trace,varargin)
+%
+% Extract event times from the ICA trace of a cell doing 2 iterations
```
kimtonyhyun: "2 iterations" of what? This comment does not explain the mathematical approach underlying this event detection script.

signal_op/extract_events.m
```diff
@@ -0,0 +1,174 @@
...
+% Hakan Inan (Mar 15)
+%
+
+%%%%
+%some parameters that can be changed within the script (advanced):
+%%%%
+%mad for df/f trace is multiplied with below to get the coarse threshold
+coarse_thresh_multiplier = 9;
+%mad for deconvolved trace is multiplied with below to get the fine threshold
+fine_thresh_multiplier = 15;
```
kimtonyhyun: Are the relative magnitudes of these two multipliers arbitrary?

signal_op/extract_candidate_events.m
```diff
@@ -1,11 +1,10 @@
-function idx_event_frames = extract_candidate_events(trace,varargin)
+function idx_event_frames = extract_candidate_events(trace,mad_scale,varargin)
 % Extract maximum points in the trace as candidate events
 % varargin : 1 optional argument. Set it to 1 if you want to plot the
 % output
```
kimtonyhyun: The display option (when the function is invoked with `varargin==1`) should provide more detail to what we're seeing, i.e. a legend (like you provide in `extract_events`).

signal_op/extract_events.m
```diff
@@ -0,0 +1,174 @@
...
+        title('raw and deconvolved ica traces together with the thresholds and the detected events')
+        legend('raw trace','deconvolved trace','coarse threshold','deconv threshold')
+    end
+end
+
+%%%%%%%%%%%%%%%%%%%
+%internal functions
+%%%%%%%%%%%%%%%%%%%
+function out = iterate_event_indices(idx_event_frames_in)
+
```
kimtonyhyun: Can you summarize what each 'iteration' does here?

signal_op/extract_events.m
```diff
@@ -0,0 +1,174 @@
...
+% Length before and after the maximum of the characteristic event, h
+num_frames_before_max = 20;
+num_frames_after_max = 100;
+
+
+idx_event_frames = extract_candidate_events(trace,coarse_thresh_multiplier,0);
+out = iterate_event_indices(idx_event_frames);
+idx_event_frames_v2 = out{1};h_1stround = out{2};
+out = iterate_event_indices(idx_event_frames_v2);
+idx_event_frames_v3 = out{1};h_2ndround = out{2};g_2ndround = out{3};
```
kimtonyhyun: "Programming" comment: The above set of iterations lends itself naturally to a for loop. You could then easily specify the number of iterations as a parameter.

signal_op/extract_events.m
```diff
@@ -0,0 +1,174 @@
...
+        figure,
+        plot(trace,'k');
+        hold on
+        plot(filtered,'m');
+        plot(1:length(trace),repmat(threshold_coarse,1,length(trace)),'--b');%coarse threshold
+        plot(1:length(trace),repmat(threshold_fine,1,length(trace)),'--r');%fine threshold
+        stem(spikes_vec_fine*threshold_fine);
+        stem(spikes_vec_coarse*threshold_coarse);
+        hold off
+        xlabel('frame')
```
kimtonyhyun: Also do `xlim([1 length(trace)])` here.

signal_op/extract_events.m
```diff
@@ -0,0 +1,174 @@
...
+    end
+end
+
+%%%%%%%%%%%%%%%%%%%
+%internal functions
+%%%%%%%%%%%%%%%%%%%
+function out = iterate_event_indices(idx_event_frames_in)
+
+    h_len = num_frames_before_max+num_frames_after_max+1;
+    %surrogate waveform
```
kimtonyhyun: Is "surrogate waveform" referring to `event_mat`, which seems to be aligned waveforms for each event?

signal_op/extract_events.m
```diff
@@ -0,0 +1,174 @@
...
+    trace_filtered = conv(trace,g,'same');
+    idx_event_frames_out = extract_candidate_events(trace_filtered,fine_thresh_multiplier,0);
+    out = {idx_event_frames_out,h,g};
+end
+
+function event_mat = construct_events(idx_event_frames)
+
+%align all the trial segments(where there is an event) and accumulate all of them.
+%The events in event_mat are of fixed length with frames before the maximum and
+%frames after the maximum determind by the function inputs.
```
kimtonyhyun: Be clearer with regards to the format of `event_mat`. Is it a matrix where each row represents an event waveform?

signal_op/extract_events.m
```diff
@@ -0,0 +1,174 @@
...
+            h(k) = sum(dum)/nonzeros;
+        end
+    end
+    %optional: smooth h before inverting
+    SGfilter = SGsmoothingfilter(20,8);
+    h = conv(h,SGfilter,'same');
+    g_len= h_len; %try to find an inverse g that has the same length as h
+    g = invert_h(h,g_len); %invert h 
+    %convolve g with the trace of the whole day and use it to calculate an
+    %adaptive threshold for detecting events
```
kimtonyhyun: There is no adaptive threshold here, only `fine_thresh_multiplier = 15`.

kimtonyhyun:
The hard-coded threshold values are not appropriate for traces that show different spiking frequencies. Here is a cell (c9m7d08 / ica001 / ic148) that was classified to be "phase-sensitive cell" but the event detection fails:
c9m7d08_ica001_ic148

Also, the deconvolved trace bares very little resemblance to the raw trace:
c9m7d08_ica001_ic148_zoom

On closer inspection, it appears that the determination of the characteristic event basically crapped out on the second iteration:
c9m7d08_ica001_ic148_char-eqn

kimtonyhyun:
The current implementation is not robust enough to be used for general event extraction.

One obvious failure mode is that the "characteristic event" function doesn't arrive at a reasonable shape in subsequent iterations (see the flat line above). At a minimum, you should detect when the characteristic event has "diverged" this way.

Or, use a simpler approach that does not involve computation of the inverse function from an empirical characteristic event function. The current approach is not numerically robust, and the central assumption (that there is a nice repeated waveform for all events of an IC) is questionable as well.

inanhkn:
The algorithm now stops iterating when the number of extracted events drop radically. We will probably end up not using deconvolution after all, but now at least it knows when to stop. 
Note: How radical the drop should be to terminate is set within the script very arbitrarily.

kimtonyhyun:
Are you going to further develop this deconvolution-based event detection scheme? If not, then we should close PR.

inanhkn:
No, I don't think I will. I'm closing it.

PR #25, opened 2015-02-13
Minor adjustments

kimtonyhyun:
@inanhkn Few minor adjustments from today:
- `view_ic_pair`: Now makes it so that the 3D view of the IC filter (in `classify_ics`) now starts off from a top-down view (see attached picture).
- `bin_movie_in_time`: The resultant movie should be floating point, since averaging multiple (integer-pixel) frames can still yield non-floating point results.
- `norm_by_background_mean`: In the special case where the quantile bounds are set to `[0 1]` (i.e. the full range) then avoid the `quantile` operation. (By the way, I've been looking at some per-frame histograms... I'm not sure that the `quantile` filtering is playing a significant effect.)

Top-down view of IC filter:
20150212_ic-filter_topdown

### DIFF
--- a/classification/support/view_ic_pair.m
+++ b/classification/support/view_ic_pair.m
@@ -18,6 +18,7 @@ daspect([1 1 z_range/max(h,w)]);
 xlabel('x [px]');
 ylabel('y [px]');
 set(gca, 'YDir', 'Reverse');
+view([0 90]);
 zlabel('IC filter [a.u.]');
 
 subplot(3,1,1);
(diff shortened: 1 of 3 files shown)


kimtonyhyun:
@inanhkn I further investigated whether the `quantile`-based filtering in the per-frame mean calculation has a significant effect.

Using c9m7d08 data, I picked random frames (#s 10, 10000, 20000), and considered the frame histogram following disk normalization. I then computed the "naive mean" (the mean over the entire image) as well as a `quantile`-filtered mean (using `[0.4 0.7]` -- the default values in the `norm_by_background_mean`). You can see that the two means are not appreciably different:
c9m7d08_frame10
c9m7d08_frame10000
c9m7d08_frame20000

However, you can see that the filtered mean is always offset to the right of the naive mean. A straightforward reason for this is that the quantile bounds of `[0.4 0.7]` is not symmetric and favors the brighter pixels. If you use a symmetric bound, e.g. `[0.3 0.7]` the two means are pretty much the same:
c9m7d08_frame10000_symmetric

Thus, my conclusion is that the quantile filtering per frame is not contributing any more than a naive mean. The extra filtering step only adds extra computation cycles.

PR #26, opened 2015-02-18
add SGsmoothingfilter

inanhkn:
@kimtonyhyun 

This is the smoothing filter I'm using for smoothing the IC traces.
Call it with the parameters `N` (`2N+1`  is the length of the filter) and `L` (desired polynomial order), e.g. `h = SGsmoothingfilter(10,8)`

### DIFF
--- a/signal_op/SGsmoothingfilter.m
+++ b/signal_op/SGsmoothingfilter.m
@@ -1,0 +1,55 @@
+function h = SGsmoothingfilter(N,L)
+% function [h,SGfilter] = SGsmoothingfilter(N,L);
+%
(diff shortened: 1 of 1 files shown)


kimtonyhyun:
Can you post an image of the effect of the filter on a trace (show the original too), for everyone's benefit?

inanhkn:
Here's one with both raw trace and median filtered trace. Compared to median filtering it seems to preserve peaks (both upward and downward) better, assuming the degree of smoothing is about the same in this particular case.

kimtonyhyun:
Looks good to me. It might be nice to have a wrapper function for applying the SG filter onto a trace.

PR #29, opened 2015-02-20
Fori2

forea:
@kimtonyhyun This function draws an outline around all the ICs in either green or red depending on if it was classified as cell or not cell by classify_ICs. Also has the option to label cells with IC# for later review.


kimtonyhyun:
It would be helpful to post a screenshot of the types of plots produced by this script (especially for Jessica, who wasn't around to see the plots in person). You can attach images to the PR discussion by dragging and dropping.

classification/view_all_classified_ics.m
```diff
@@ -0,0 +1,58 @@
...
+% 2015 02 19 Fori Wang
+
+figure;
+
+% load ICs
+load(ica_mat);
+
+% Pull out classification data
+fid = fopen(class_file);
+results = textscan(fid,'%s %s %s %s');
```
kimtonyhyun: The classification file a comma-separated file currently takes the form:
```
...
37, cell
38, cell
39, not a cell
40, cell
...
```

So, it's not correct to scan this file by `textscan(fid,'%s %s %s %s')`, which means to look for four columns of strings per line.

What you should do is to use the existing function [`load_classification(...)`](https://github.com/inanhkn/analysis/blob/master/classification/load_classification.m) for reading the classification file.

forea:
Here's an example plot from this function

example_view_all_classified_ics

classification/view_all_classified_ics.m
forea: @kimtonyhyun changed to load_classification(...)

kimtonyhyun: After changing the code, you need to re-push to the branch to update the PR.

(I'm bringing the postdoc candidate from Clark.)

kimtonyhyun:
On the title, I see that you are getting subscripts erroneously due to the underscore characters. You can avoid this, for the `pwd` for example, by: `strrep(pwd, '_', '\_')`.

forea:
thanks. forgot to commit before pushing.

also added strrep... was wondering if there was a way to fix that.

kimtonyhyun:
Looks good.

PR #30, opened 2015-02-23
Revisions to disk normalization

jmaxey:
This revision removes mean normalization and reads in a single frame at a time rather than loading a full movie into memory.

@kimtonyhyun  @inanhkn

### DIFF
--- a/image_op/norm_by_disk_filter.m
+++ b/image_op/norm_by_disk_filter.m
@@ -1,11 +1,14 @@
-function norm_by_disk_filter(hdf5Dir,hdf5Name,dataset,varargin)
+function norm_by_disk_filter(inputDir,inputName,inputDataset,outputDir,outputName,outputDataset,varargin)
 %Normalize every frame in the movie by a disk filtered version of itself
 %
-%   movie: movie matrix, [h x w x num_frames]
+%   inputDir, inputName, inputDataset = Path, name, and dataset name of the
+%   input hdf5 movie
+%   outputDir, outputName, outputDataset = Path, name, and dataset name of
+%   the output hdf5 movie
 %   disk_radius may be passed as an input argument. Default for disk_radius
 %   is 15.
-% 2015 01 31 Tony Hyun Kim (Latest Revision: Jessica Maxey, 15-Feb-23)
 %
+% 2015 01 31 Tony Hyun Kim (Latest Revision: Jessica Maxey, 15-Feb-23)
 
 if isempty(varargin)
     disk_radius = 15;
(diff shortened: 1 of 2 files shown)


inanhkn:
Is there a reason you write the normalized movie into the same hdf5?

jmaxey:
I find it easier to have all of the videos stored in a single file rather than multiple files from the same mouse/day. However, I will change the file so that the user can specify both different dir, name, and datasets for the input and output files.

inanhkn:
I think it is actually more difficult in some aspects. A lot of times I pull single movies file from other computers in the lab over the network, I would find it very annoying to have to transfer a 3x bigger file for only pulling the df/f movie :). Can you still keep it the way it is, i.e. separate files for each?

inanhkn:
Unless there is a specific reason for doing it the way you want to do of course

jmaxey:
It's no problem to let the user decide how they want to store the new file. I'll update the PR shortly. It will allow both separate and single file storage options.

inanhkn:
Awesome :) Can you make the function so that it generates a separate file by default ?

analysis
```diff
@@ -0,0 +1 @@
+Subproject commit 47c80b5d1edf8f9543810cda7065aa835dddff7e
```
kimtonyhyun: What's this file?

inanhkn:
It looks like this script is basically going to read from a specific path (`/Data/Images`) and write again to a specific one (`/Data/Norm`) if the user decides to store the normalized movie in the same file.The same arguments are valid for the case of different input and output files (both would be, say, (`/Data/Images` in that case). So, I think having `dataset` as input might be a little redundant here. It would be a better option to have same file versus different files as an input parameter. The function can write to the same file if no `outputDir,outputName` arguments are input and write to a different file specified by these parameters otherwise.

jmaxey:
I think we should decide on dataset names. Later in the pipeline, dataset names are being hardcoded (I just changed this to work with my file). It would make since to have a standard set of names. I like the single file format, however, we cannot have the same dataset name for each step in this case. Could we have different dataset names for each step of the process regardless if they are the only dataset in a file? 

I just committed the revisions to the HDF5 concatenation file and it was tacked onto this pull request. Is there a way to separate these?

kimtonyhyun:
Look into `git reset --soft HEAD~n` for going back to a previous commit in a branch -- to separate the PR from the concatenation fix.

kimtonyhyun:
Or, just go to a fresh branch from `master`, and reimplement the HDF5 fix in the fresh branch. We can resolve the merge conflict later.

jmaxey:
I'll do the second option. Not sure how to do the first.

----- Original Message -----

kimtonyhyun:
I'm also in favor of separate files for movies at different points of the processing. At Herrin, we're working on multiple computers where we have to copy over movies. Given that the DFF movie is basically all we need to transport at the end of the day, it doesn't make sense to move around 100+ GB files that contain the history of the movie. It doesn't make sense in terms of copy times, and also because the destination HDDs are not all 4 TB drives.

As for the HDF5 directory names, I don't want to have to specify the dataset name explicitly whenever I need to load a movie from file. If multiple names are going to be used internally, then the load/save wrappers (e.g. `load_movie_from_hdf5`) should take care of it behind the scenes.

What is the argument for having multiple directory names?

jmaxey:
It's mostly for storing image stacks in the same file. If everyone is against this, I can hard code '/Data/Images' as the directory for all files.

----- Original Message -----

kimtonyhyun:
This PR is out of date given the fix to `concatenateHDF5`. I'll create a new PR with the file-to-file disk normalization function, but one that works with the convention that there is one movie per HDF5 file.

PR #31, opened 2015-02-24
add extract_candidate_events.m

inanhkn:
@kimtonyhyun  @forea @jmaxey 
The file `extract_candidate_events.m` takes in the ica trace and outputs the indices in the trace where there are local maxima (candidate events). 
Call the script with an additional argument set to `1` if you want to see the trace together with the points corresponding to the candidate events.

### DIFF
--- a/signal_op/calc_localmax_above_threshold.m
+++ b/signal_op/calc_localmax_above_threshold.m
@@ -1,0 +1,28 @@
+function localmax = calc_localmax_above_threshold(threshold,data)
+%function calc_localmax_above_threshold(threshold,data)
+%
+%find the local maxima in a sequence of points provided that the maxima are 
+%above a specified threshold
+%
+%Hakan Inan (Jan 15)
+%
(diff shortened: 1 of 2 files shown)


kimtonyhyun:
Do you envision that `calc_localmax_above_threshold` will be used for functions other than `extract_candidate_events`? If not, you should make `calc_...` an internal function to `extract_...`

inanhkn:
Actually, `extract_events.m` , which uses deconvolution routine to extract events also calls that function. At this point I don't know if we will be using `extract_events.m` , but I think we should still keep `calc_localmax_above_threshold.m` as a separate file for now.

kimtonyhyun:
Looks good to me as a first pass. However, everyone should be aware that there are some events that still may be missed by this method (i.e. still room for future improvement).

Here's an example from c09m7d07:
20150225_c9m7d07_ic014
20150225_c9m7d07_ic140

PR #33, opened 2015-02-27
Trim behavior video

kimtonyhyun:
@forea 

The function `trim_behavior_video(plusmaze_source, behavior_source, trim)` will generate a new behavior video that matches the concatenated Miniscope video (i.e. produced by `concatenateHDF5` and `concatenate_bigtiff`) frame-by-frame. The `trim` parameter needs to match those used in the Miniscope concatenation, of course. (We've been using `trim = [15 5]` so far.)

Note! The output "trimmed" behavior video will have the number of frames according the plus maze output. If the behavior video has dropped frames, then the behavior video will be "uniformly stretched" to match the plus maze frame counts. (We should be able to improve on the localization of dropped behavior frames in the future.)

Performance-wise, it took about 30 min for my laptop to process c9m7d07 behavior video. It may be possible to improve the performance by batching the writes to the output file, which right now occurs frame-by-frame.

### DIFF
--- a/io/trim_behavior_video.m
+++ b/io/trim_behavior_video.m
@@ -20,11 +20,13 @@ function trim_behavior_video(plusmaze_source, behavior_source, trim)
 % Get the trial frames (beginning and end) according to plusmaze output and
 % trim the beginning and end
 frame_indices = get_trial_frame_indices(plusmaze_source);
-trial_frame_indices = [frame_indices(:,1)+trim(1) frame_indices(:,4)-trim(2)];
-num_trials = size(trial_frame_indices,1);
-num_trial_frames = trial_frame_indices(num_trials,2); % Very last frame
+num_trials = size(frame_indices,1);
+num_trial_frames = frame_indices(num_trials,4); % Very last frame
 fprintf('  PlusMaze output (%s) has %d frames\n',...
     plusmaze_source, num_trial_frames);
+
+% Frames to keep
+trial_frame_indices = [frame_indices(:,1)+trim(1) frame_indices(:,4)-trim(2)];
 
 % Expected number of trimmed frames
 num_trimmed_frames = sum(diff(trial_frame_indices,1,2)+1);
--- a/io/trim_behavior_video.m
+++ b/io/trim_behavior_video.m
@@ -1,0 +1,71 @@
+function trim_behavior_video(plusmaze_source, behavior_source, trim)
+% Extract the trial frames of the behavior video, so that the resulting
+%   video lines up with the concatenated Miniscope movie (e.g. as produced
+%   by `concatenateHDF5` or `concatenate_bigtiff`). The `trim` parameter
+%   needs to match the values used in the Miniscope concatenation.
+%
+% Inputs:
+%   plusmaze_source: Text file output from the plus maze
+%   behavior_source: Behavior video (MPEG-4)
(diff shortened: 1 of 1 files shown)


kimtonyhyun:
FYI, I did some manual inspection of the behavior dropped frames.

The c9m7d07 behavior video has 37 dropped frames. In trial 1, there are 7 missing frames. This would imply that dropped frames tend to group up.

kimtonyhyun:
Annoyingly, Matlab on Linux is having difficulties reading in MPEG-4 movies via `VideoReader`. (Appears to be codec issue.) @forea How were you reading in the behavior video in Linux previously?

The code should run on Windows.

inanhkn:
@kimtonyhyun the codec issue is  due to an incompatibility with  Ubuntu 14.10. It needs a different kind of codec than the previous versions. Fori and I dealt with it by adding to the repository a PPA specific to Ubuntu 14.10. I believe it was this one: `ppa:mc3man/gstffmpeg-keep`

inanhkn:
By the way, did you just manually go to the frame index in th behavior video where the first trial seemed to finish to check the number of dropped frames?

kimtonyhyun:
Cool. I'll check out `papa:mc3man/gstffmpeg-keep`. However, Fori had some residual issues with MP4 files even on her Ubuntu machine (it would read an incorrect number of frames); the workaround for now was to use the Windows machines. It'll be good to get the Ubuntu/MP4 issue resolved for good though.

Yes, I estimated the number of dropped frames in trial 1 by manually checking the frames near the Trial1/2 boundary. We are intending on a way to do this automatically (to look for trial boundaries based on the frames of the behavior video), so that we don't have to "smear" the missing frames across all trials, but only within the trials where there are dropped frames. In retrospect, it could be very important to not "leak" across trials. For example, an event at the end of one trial could mistakenly show up as the beginning of the next, which would screw up our spatial interpretation of the cell firing.

inanhkn:
Yeah, I agree. This might be done by tracking the mouse and looking for sustained discontinuities in the mouse position. Or, the algorithm might only only look for position switches when the mouse has reached one end of the maze (suggesting he's going to start from a different position shortly)

kimtonyhyun:
Yes, my plan was to try the discontinuity approach,

-Tony (from phone)

forea:
I have a script for this that I will include in my mouse tracking PR
I'd like to test it first on a cropped behavior video

kimtonyhyun:
Here's a link to a cropped version of the c9m7d07 behavior vide: https://www.dropbox.com/s/zkaa8lpt1g4c9fk/mouse7_day07_ego-left_cropped.mp4?dl=0

(I'll send the PR for cropping shortly.)

forea:
thanks, trying it out now

kimtonyhyun:
Actually, use this one (still c9m7d07). The cropped video was generated with less compression: https://www.dropbox.com/s/zkaa8lpt1g4c9fk/mouse7_day07_ego-left_cropped.mp4?dl=0

kimtonyhyun:
Closing for now -- will be replaced with a version that does not leak frames across trials.

forea:
This is cropped but not trimmed, right?

kimtonyhyun:
The Dropbox link is for the cropped video (not trimmed)

PR #35, opened 2015-02-28
Kimth/time bin indices

kimtonyhyun:
@forea 

Added new method `bin_frame_indices` that will apply temporal binning to trial frame indices, as to be compatible with the 2x time-binned movies we've been using lately.

We can indicate to `classify_ics` that temporal binning has been applied through a new field `sources.time_bin`. This allows `classify_ics` to display the IC traces as a trial superposition (as it has done in the past).

I also refactored `classify_ics` to use the `compute_ic_boundary` function (introduced in the alignment PR). The manual center-of-mass calculation was replaced with a Matlab built-in (`regionprops`) as well.


jmaxey:
It is unclear what the input variable "frame_indices" is. Is this a matrix containing the frames for the start and stop of each trial? Does it only handle start and stop and not gate up and gate down frames? 

----- Original Message -----

kimtonyhyun:
Good point. I should make the comments a bit more explicit about the structure of `frame_indices`. I will do so in the next PR.

To answer your question though, it actually will work with `[start stop]` format or `[start gate-up gate-down stop]` format, though I had intended the latter.

PR #38, opened 2015-03-01
Crop behavior video

kimtonyhyun:
@forea 

This is the script I used to generate the cropped behavior movies.

When you run the script (e.g. `crop_behavior_video(mouse07_day07_ego-left.m4v)` -- by the way `.m4v` and `mp4` are equivalent extensions), the script will prompt you to draw a rectangular region over the movie. When you are finished manipulating the rectangle, then double click _inside the rectangle_ to proceed:
crop-rect

It took about 15 min on my laptop to crop out c9m7d07.

I noticed that you have a helper script in your `Fori2` branch that pulls in chunks of the movie at a time, instead of frame-by-frame. Once that script is pulled in (and cleaned up a bit), then I'd like to update this function to process chunks of movie at a time.

### DIFF
--- a/behavior/crop_behavior_video.m
+++ b/behavior/crop_behavior_video.m
@@ -1,0 +1,61 @@
+function crop_behavior_video(behavior_source)
+% Spatially crop the behavior video. The cropping region is selected
+%   interactively during the script. Writes the resulting file as MPEG-4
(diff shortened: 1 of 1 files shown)


kimtonyhyun:
Wow, it's March!

PR #39, opened 2015-03-02
Fixed bug in concatenation file

jmaxey:
Fixed the frame count error in the file.

@kimtonyhyun

### DIFF
--- a/io/concatenateHDF5.m
+++ b/io/concatenateHDF5.m
@@ -101,6 +101,7 @@ for i=1:length(list)
         
         %%% Replace any dropped frames with the frame immediately
         %%% preceeding it
+        cropDroppedFrames = 0;
         if(droppedFrames ~= 0)
             for j=1:length(droppedFrames)
                 disp('DroppedFrame');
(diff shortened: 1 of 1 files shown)


kimtonyhyun:
Checking it out

kimtonyhyun:
I just tried out this code on c9m7d09 (which has a non-glitched PlusMaze text file). I expected the concatenated (and trimmed) movie to have 46228 frames, but found only 46226 in the resulting HDF5. Is the current code correctly writing in the dropped frames into the HDF5?

jmaxey:
Let me check. I may need to alter something else since the dropped frames are in the first cropped 15. 

----- Original Message -----

kimtonyhyun:
Same on c9m7d06, the concatenated result is 2 frames shorter than expected.

kimtonyhyun:
Does `concatenateHDF5` assume that the first file is always a valid trial? This assumption may be false in some cases, if we needed to do a repeat on the first trial.

jmaxey:
I know what's wrong and am working to fix it now. I will make sure that dropping the first frame is not a problem.

----- Original Message -----

jmaxey:
I have to go to office hours soon and then have class but will keep working on it. Sorry for the delay.

----- Original Message -----

kimtonyhyun:
Outdated -- closing.

PR #41, opened 2015-03-03
Fori2

forea:


forea:
@kimtonyhyun @inanhkn @jmaxey 
This PR includes the following:
- get_mouse_XY_pos.m: Get mouse's position from (cropped) behavior video
- get_mouse_XY_pos_live_monitor: Same as above, but lets user watch the position extraction
- 3 helper files for above: findMouse.m, make_frame_indices.m, and centroid_plotter.m
* note these functions only work on Windows because of weird read errors on Linux*

within "FindTrials": these functions are used to find start and end of trials (since some behavior videos have dropped frames, this is useful for isolating dropped frames to within a specific trial)
- find_start_end_of_trials.m: Using position data, find the start/end of trials (looks for jumps)
- mouse_path_plotter.m: This is a useful function to visualize all the mouse paths that above function has found. Note that probe trials will cause some bleed-through (3 trials combined into 1);
- scrollsubplot: someone else's function that helps to make scrollable plots

forea:
Example screenshot of what mouse_path_plotter.m produces (using position data from get_mouse_XY_pos.m and trial start/ends from find_start_end_of_trials)
screenshot from 2015-03-02 18 42 18

forea:
Also, using find_start_end_of_trials.m, I can compare the trial start/ends that the behavior video has to the plusmaze txt file to isolate behavior dropped frames to specific trials. I have done this for days 7 and 8 as a test. This was a good sanity check, as the # dropped frames I found matched the Cohort 9 Spreadsheet (37 and 28 dropped frames, respectively).

This txt file (saved in cohort9 DB) summarizes dropped frames: https://www.dropbox.com/s/tjsbsz902ae6xvs/mouse7_day07_ego-left_dropped_behavior_frames.txt?dl=0. >>>> I imagine we can use this in the "trim_behavior_video" function to smear within trials. Then, recalculate position data based on smeared video so that each miniscope frame has position data.

This ods (fake excel) file is a record of how I (manually) found all the dropped frames: https://www.dropbox.com/s/1ag8ikyglay9ski/mouse7_day07_ego-left_dropped_frames.ods?dl=0

forea:
(there are equivalent txt and ods files for day08 too)

kimtonyhyun:
Can you move the `scrollsubplot` into the `utils` directory? It may be useful for other functions.

Also, can you rename `make_frame_indices` into something like `make_frame_chunks`? We are overloading the term `frame_indices`. It would be good to denote the output of this function as "frame_chunks" to indicate that this is a helper function for loading chunks of the movie at a time, and is not producing the frame indices corresponding to trial phase. Also, can you move this file into `io`, since I think it will be generally useful for other methods as well (other functions that process chunks of the movie at a time).

forea:
@kimtonyhyun good ideas. Done!

io/make_frame_chunks.m
```diff
@@ -0,0 +1,24 @@
...
+
+    frame_chunks = [];
+    for chunk_idx = 1:chunk_size:num_frames
+        if chunk_idx+chunk_size-1 < num_frames
+            frame_chunks = [frame_chunks; chunk_idx chunk_idx+chunk_size-1];
+        else % for the last chunk        
+            frame_chunks = [frame_chunks; chunk_idx num_frames];
+        end
+    end
+
```
kimtonyhyun: Not critical, but here is a more efficient way to compute `frame_chunks` without looping:
```
x = (1:chunk_size:num_frames)';
y = [x(2:end)-1; num_frames];
frame_chunks = [x y];
```

behavior/MouseTracking/get_mouse_XY_pos.m
```diff
@@ -0,0 +1,75 @@
...
+% 2015-02-26 Fori Wang
+
+    behavior_vid = VideoReader(movie);
+    num_frames = behavior_vid.NumberofFrames;
+
+    % initialize centroids
+    centroids = cell(num_frames,1);
+    
+    % setup chunks of frames to read in movie
+    frame_indices = make_frame_indices(num_frames,1000);
```
kimtonyhyun: Now that you've renamed `make_frame_indices` into `make_frame_chunks`, you need to update these functions as well. While you are at it, can you replace `frame_indices` with `frame_chunks`?

behavior/MouseTracking/get_mouse_XY_pos.m
```diff
@@ -0,0 +1,75 @@
...
+    bg_image = mean(bg_vid,4);
+    bg_image = uint8(bg_image);
+    
+    for idx = 1:length(frame_indices)
+
+        fprintf('  Processing frames %d to %d (%s)\n',...
+            frame_indices(idx,1), frame_indices(idx,2), datestr(clock));
+
+        % read in all of the frames for the trial at the beginning
+        frame_range = [frame_indices(idx,1) frame_indices(idx,2)];
```
kimtonyhyun: Isn't this just `frame_indices(idx,:)`?

behavior/MouseTracking/get_mouse_XY_pos.m
```diff
@@ -0,0 +1,75 @@
...
+        video = read(behavior_vid,frame_range);
+        
+        c_old = [0 0];
+
+        for frame_idx = frame_indices(idx,1)-(frame_indices(idx,1)-1):...
+                        frame_indices(idx,2)-(frame_indices(idx,1)-1)
```
kimtonyhyun: Is this just `frame_idx = 1:...`?

kimtonyhyun: It may be conceptually clear to write: `for frame_idx = 1:size(video,4)`

behavior/MouseTracking/get_mouse_XY_pos.m
```diff
@@ -0,0 +1,75 @@
...
+           
+           % Update original image CData
+            image = video(:,:,:,frame_idx);
```
kimtonyhyun: Convert to `rgb2gray`?

kimtonyhyun: Doing so could speed up your code significantly, since it reduces all the frames from `[height width RGB]` to just `[height width]`. I see that `findMouse` has some nontrivial image operations in it.

behavior/MouseTracking/get_mouse_XY_pos.m
```diff
@@ -0,0 +1,75 @@
...
+    for idx = 1:length(frame_indices)
+
+        fprintf('  Processing frames %d to %d (%s)\n',...
+            frame_indices(idx,1), frame_indices(idx,2), datestr(clock));
+
+        % read in all of the frames for the trial at the beginning
+        frame_range = [frame_indices(idx,1) frame_indices(idx,2)];
+        video = read(behavior_vid,frame_range);
+        
+        c_old = [0 0];
```
kimtonyhyun: `c_old = [0 0]` should be outside of the `idx` loop. Otherwise, if you lose the position of the mouse at the first frame of a chunk, then you will not be able to use the previous known position of the mouse.

behavior/MouseTracking/get_mouse_XY_pos_live_monitor.m
```diff
@@ -0,0 +1,112 @@
...
+
+                % Update subplots
+                set(k,'XData',c_new(1),'YData',c_new(2),'color','b');
+                set(l,'XData',c_new(1),'YData',c_new(2),'color','b');
+                
+            end
+        end    
+    end
+    
+    centroids=cell2mat(centroids); %convert to matrix
```
kimtonyhyun: It looks like `centroids` can be a matrix to begin with?

behavior/MouseTracking/findMouse.m
```diff
@@ -0,0 +1,40 @@
...
+       
+       % get rid of smaller dots that are not the mouse
+       final_image = bwareaopen(sub_image,200,8);          
+       
+       % now fill in the blob
+       se = strel('disk',5);
+       final_image = imdilate(final_image,se);
+            
+        % get centroid, area, and length of blobs
+        blobs = regionprops(final_image(:,:,1), 'area','centroid','minoraxislength');
```
kimtonyhyun: Indentation is offset on the above two lines.

behavior/MouseTracking/centroid_plotter.m
```diff
@@ -0,0 +1,22 @@
+function centroid_plotter( movie, centroids_mat )
+% Plots all centroids on first frame of behavior movie for general inspection
+% 
+% Input:
+%     Behavior movie
+%     centroids generated by get_mouse_XY_pos function
+%
+% Note: Runs sort of slow because you have to access each cell one by one
+% via for loop
```
kimtonyhyun: Outdated comment?

behavior/MouseTracking/findMouse.m
```diff
@@ -0,0 +1,40 @@
...
+              
+       % isolate mouse by subtracting frame from background image
+       sub_image = bg_image-actual_image;
+       
+       % get rid of smaller dots that are not the mouse
+       final_image = bwareaopen(sub_image,200,8);          
+       
+       % now fill in the blob
+       se = strel('disk',5);
+       final_image = imdilate(final_image,se);
```
kimtonyhyun: Just a note: For future optimization, if we need it, we should try to see if we can be "leaner" in the image processing steps in `findMouse`.

behavior/MouseTracking/FindTrials/find_start_end_of_trials.m
```diff
@@ -0,0 +1,105 @@
...
+% will jump from e.g., south arm to east arm)
+%
+%   Note: this function is used by mouse_path_plotter
+% 
+% Input:
+%     centroids.mat file from get_mouse_XY_pos
+% 
+% Returns
+%     trial_frames: Array of start and end frames [start end; start end; ...]
+%     tracking_error_frames: Frames that are not start/end put show a jump
```
kimtonyhyun: Grammar: "put show a jump". Did you mean "but"?

behavior/MouseTracking/FindTrials/find_start_end_of_trials.m
```diff
@@ -0,0 +1,105 @@
+function [ trial_frames, tracking_error_frames] = find_start_end_of_trials( centroids_mat )
+% Using position data from get_mouse_XY_pos (using behavior vid), find
+% frames where trial ends and next trial starts (based on fact that mouse
+% will jump from e.g., south arm to east arm)
```
kimtonyhyun: What happens for trials that go like: east-->south, south-->west, west-->north (e.g. on ego-left trials with probe trial in the middle)? 

Is the output of this function intended to be a "suggestion" for the trial boundaries -- i.e. that the user should manually inspect the output? It's fine if so, but it should be explicitly stated.

kimtonyhyun:
Left some feedback for the PR. By the way, it seems to me that the two features of the PR are conceptually separate. One is the simple mouse tracking (which generates a list of centroids from the video); and the other is the detection of the trial boundaries from the behavior video (which operates from the list of centroids).

behavior/MouseTracking/FindTrials/mouse_path_plotter.m
```diff
@@ -0,0 +1,42 @@
...
+% 
+% Input:
+%     movie: Behavior video
+%     centroids_mat: position data from get_mouse_XY_pos
+%
+% 
+% 2015-03-01 Fori Wang
+
+    % find trial start and ends
+    [trial_frames,~]=find_start_end_of_trials(centroids_mat);
```
kimtonyhyun: This function should allow the user to specify a PlusMaze text file. If the behavior video and the text file are aligned, then the trial frame indices from the text file provide perfect knowledge into the trial boundaries. The function should try to recover trial frames through `find_start_end_of_trials` only if the plus maze text file is not provided.

kimtonyhyun:
The format of behavior dropped frame tables as in `mouse7_day07_ego-left_dropped_behavior_frames.txt` looks good to me. I'll generate a "trial boundary preserved smeared" version of the behavior video for c9m7d07.

behavior/MouseTracking/get_mouse_XY_pos.m
```diff
@@ -0,0 +1,75 @@
...
+    num_frames = behavior_vid.NumberofFrames;
+
+    % initialize centroids
+    centroids = cell(num_frames,1);
+    
+    % setup chunks of frames to read in movie
+    frame_indices = make_frame_indices(num_frames,1000);
+    
+    % setup background image: average of 5000 frames
+    bg_vid = read(behavior_vid,[1 5000]);
```
kimtonyhyun: The video is grayscale, but each frame of `bg_vid` is represented as RGB. An efficient way to convert to grayscale is to write:
```
bg_vid = read(behavior_vid, [1 5000]);
bg_vid = squeeze(bg_vid(:,:,1,:));
```
This keeps only the red channel of the RGB video. (For a true grayscale video, the R, G, and B values are identical.) At the end, `bg_vid` has the dimensions `[height x width x num_frames]` (i.e. 3D movie stack, not 4D).

kimtonyhyun: This should also be performed for `video = read(behavior_vid, frame_range)` in the chunk loop below.

behavior/MouseTracking/get_mouse_XY_pos_live_monitor.m
```diff
@@ -0,0 +1,112 @@
+function [ centroids ] = get_mouse_XY_pos_live_monitor( movie )
```
kimtonyhyun: Generally speaking, one should not duplicate functions like `get_mouse_XY_pos` and `get_mouse_XY_pos_live_monitor`. Rather, `get_mouse_XY_pos` should have a user-specified flag that indicates the output intermediate results or not.

behavior/MouseTracking/findMouse.m
forea: Yes, that would be good. I think bwareaopen is slow (anecdotally).

behavior/MouseTracking/FindTrials/find_start_end_of_trials.m
forea: Yes, just a suggestion; has been added to comments.

forea:
Major changes (in addition to minor bug fixes)
- Combined get_mouse_XY_pos and get_mouse_XY_pos_live_monitor into one function (get_mouse_XY_pos), with optional input 'displayTracking'
- mouse_path_plotter now uses plusmaze data for trial frame info, and if not given, will generate trial frames from behavior video

Not changed (for now)
- RGB 4D video. For now, switching back to 4D. For some reason position detection is not as good in 3D mode, and haven't figured out why yet. 4D works, it's just slow.

kimtonyhyun:
Let me know once you've made a final pass through the code.

forea:
@kimtonyhyun made the final pass

behavior/MouseTracking/get_mouse_XY_pos.m
```diff
@@ -0,0 +1,144 @@
...
+                c_new = s(id(1)).Centroid(1:2);
+                centroids(frame_chunks(idx,1)+frame_idx-1,:) = c_new;
+                c_old = c_new;
+                
+                if display_tracking
+                    % Update subplots
+                    set(k,'XData',c_new(1),'YData',c_new(2),'color','b');
+                    set(l,'XData',c_new(1),'YData',c_new(2),'color','b');
+                    pause(0.00001);
+                end
```
kimtonyhyun: If you pull out this code chunk `if display_tracking ... end` out of the containing `if isempty(area_vector) ... end`, then you don't have to repeat the display block.

PR #43, opened 2015-03-03
Kimth/trim behavior

kimtonyhyun:
@forea 

Function to trim the behavior video (to match the Miniscope concatenation -- via `concatenateHDF5`). The dropped frames of the behavior video is "smeared" at the trial level (i.e. no bleedthrough from trial-to-trial).

Example usage with c9m7d07 as an example:
```
% Load the file containing dropped frames info
dropped_table = load('mouse7_day07_ego-left_dropped_behavior_frames.txt');

trim_behavior_video('mouse7_day07_ego-left.txt', 'mouse7_day07_ego-left_cropped.mp4', [15 5], dropped_table);
```
The function generates a new MPEG-4 video with `_trim15-5` (for `trim=[15 5]`) concatenated to the name.

(Note `dropped_table` is an optional input, and may be omitted if the behavior and text files sync up.)


behavior/trim_behavior_video.m
```diff
@@ -0,0 +1,124 @@
...
+fprintf('  Behavior video (%s) has %d frames\n', behavior_source, num_behavior_frames);
+
+assert(num_behavior_frames == dropped_frame_indices(end,2),...
+       '  Dropped frame table is inconsistent with actual number of frames in behavior movie!');
+
+% We don't handle the case where the number of frames in the behavior video
+%   exceeds the frame count indicated by the PlusMaze
+num_actual_dropped_frames = num_frames - num_behavior_frames;
+assert(num_actual_dropped_frames >= 0,...
+       '  Behavior video has more frames (%d) than indicated by PlusMaze output (%d)!',...
```
forea: should this be 'Behavior video has fewer frames..." ?

kimtonyhyun: It's correct as written.

Matlab's `assert` makes sure that the first condition (`num_actual_dropped_frames >= 0`) is true. The text is only displayed in the case that the assertion fails (i.e. the case `num_actual_dropped_frames < 0`)

forea:
Add a note that this function should run on Windows?

I tried it out on Linux just to check, it stops after hitting "Dropped frame table is inconsistent with actual number of frames in behavior movie!" (in this case VideoReader left out 2 frames).

kimtonyhyun:
Well, the issue with Linux is that the MPEG-4 videos yield incorrect number of frames as you noted. It's not specific to this function but is general across all behavior video processing, so I don't think the note is necessary.

PR #44, opened 2015-03-04
Added varargout "trial_frames" generated by find_start_end_of_trials

forea:
Small change to mouse_path_plotter code to also spit out trial_frames from behavior movie if no plusmaze data was provided (i.e. function had to figure out trial_frames on its own)

### DIFF
--- a/behavior/MouseTracking/FindTrials/mouse_path_plotter.m
+++ b/behavior/MouseTracking/FindTrials/mouse_path_plotter.m
@@ -33,7 +33,7 @@ function [trial_frames] = mouse_path_plotter( movie, centroids_mat, varargin)
     
     figure;
     
-    for trial_idx = 1:length(trial_frames)
+    for trial_idx = 1:size(trial_frames,1)
         trial_start = trial_frames(trial_idx,1);
         trial_end = trial_frames(trial_idx,2);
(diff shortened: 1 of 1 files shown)


kimtonyhyun:
How about always outputting `trial_frames`, rather than using `varargout`? (The calling function doesn't always have to receive the output.)

forea:
Cool, I didn't know that was possible. Updated!

kimtonyhyun:
By the way, does `trial_idx = 1:length(trial_frames)` work? Since `trial_frames` is `[n x 2]` I thought that one needs to write `trial_idx = 1:size(trial_frames,1)`

forea:
It works. L = length(X) returns the length of the largest array dimension in X... so its possible if there were only 1 trial it would be wrong. I guess size(trial_frames,1) is more accurate.

http://www.mathworks.com/help/matlab/ref/length.html.

kimtonyhyun:
We should write `size(trial_frames,1)` in cases like this -- to err on the side of explicit. Otherwise (though probably not in this case), we may run into hard-to-debug scenarios.

forea:
Yep, updated

kimtonyhyun:
Looks good to me.

PR #45, opened 2015-03-05
Kimth/trial aware binning

kimtonyhyun:
As per our email discussion, our existing movie binning method had the problem that frames could be averaged across the trial boundary. The updated `bin_movie_in_time` method now takes trial boundaries into account when performing temporal binning, so that there is no bleed-through across trials.

In this implementation, the binning operation takes place entirely in memory (both the input and output movies are in memory). I will follow up with file-to-file version of the code.

Some other minor, but important remarks:
- `get_trial_frame_indices` now returns the frame indices cast as `double` type, rather than (previously) `int32`. This is so that we can avoid quirks with integer arithmetic (e.g. `int32(3)/2 = 2` in Matlab). With this change, I don't need to cast the frame indices into `double`s in the calling function (as in `trim_behavior_video`).
- `io/bin_frame_indices` is, for the time being, _incompatible_ with the (trial-boundary preserving) binned movie generated by `bin_movie_in_time`. The reason is that `io/bin_frame_indices` is still being used by some movies that were binned by the previous method (e.g. Jessica and Day 26). Once we have phased out the previously binned movies from circulation, then we can update `bin_frame_indices`.

@forea I also added two methods (`bin_behavior_video` and `subsample_behavior_video`) to perform the same trial-boundary aware binning on the behavior videos. The difference between the two is that: `bin_behavior_video` performs frame binning by averaging frames, whereas `subsample` picks 1 out of every `bin_factor` frames.


image_op/bin_movie_in_time.m
```diff
@@ -1,26 +1,81 @@
-function M_b = bin_movie_in_time(M, bin_factor)
```
kimtonyhyun: Note, basically the entire file has been rewritten. It may be easier to review the file by clicking the `View` button associated with this file, rather than try to read the line-by-line diff.

kimtonyhyun:
A note regarding the two different methods for time binning the behavior movie.

My first implementation of the behavior time binning was to average the frames, just like the Miniscope movie. You can see the result here: [c9m7d08_ti2-mean.mp4](https://www.dropbox.com/s/l5gdh4dp1nvkg2g/c9m7d08_ti2-mean.mp4?dl=0). (Download to avoid artifacts.) In general, I noticed that averaging frames led to seeing "doubles".

My second implementation was then to subsample every `bin_factor` frame, and avoid averaging of the behavior video. The reuslt is here: [c9m7d08_ti2-sub.mp4](https://www.dropbox.com/s/8g1ykbbrfaed067/c9m7d08_ti2-sub.mp4?dl=0).

Overall it seemed to me that, for the behavior video, subsampling the frames was preferable to taking an average of multiple frames. Thoughts?

kimtonyhyun:
One final thought. It seems to me that the processed behavior videos (cropped, trimmed, subsampled) show significantly more compression artifacts (presumable MPEG-4) than the original file. Can you guys confirm?

kimtonyhyun:
@forea Let me know how you feel about the quality of the behavior movies after processing in Matlab. Some other workflows:
- Convert the MPEG-4 from the beginning into a full-blown HDF5 from the beginning (like the Miniscope recordings). This would avoid compounding encoding artifacts by repeated generation of MPEG-4 movies. Note also that this would make the behavior videos compatible with Linux. 
-  Generate the `centroids.mat` (mouse path) from the full behavior video, then parse that down via trial frame indices -- rather than parsing the behavior movie itself.

forea:
I'll check out the quality, but i do like the idea of a Linux compatible
behavior video...!

forea:
I hit the following error on the last trial (110) after trying out
trial_indices = get_trial_frame_indices('mouse7_day07_ego-left.txt');
subsample_behavior_video('mouse7_day07_ego-left_cropped.mp4',2,trial_indices);

========
Error using VideoReader/read (line 145)
Frame Range must be a 1x2 vector of frame indexes with the first value less
than or equal to the second value.
A Frame index must be a numeric value greater than zero and less than or equal
to the number of frames in the file.

Error in subsample_behavior_video (line 35)
    movie = read(behavior_video,...

forea:
nevermind! forgot to use `compress_frame_indices` first

forea:
I've successfully run:
`trim_behavior_video`
`subsample_behavior_video`
`bin_frame_indices2`
`get_trial_frame_indices`

jmaxey:
It looks good. Why do you multiply by the bin_factor rather than just using it as your increment amount in the loop? (e.g. k=1:bin_factor:binned_frames_per_trial(trial_idx)) This is just a personal preference in indexing. I'll go ahead and merge it.

jmaxey:
Also, how do you ensure that all frames are included even if the number of frames in the trial is not evenly divisible by the bin_factor?

kimtonyhyun:
Regarding multiplying by `bin_factor`, can you make the comment directly on the line of the code, so that it's easier to see what you're referring to?

behavior/bin_behavior_video.m
```diff
@@ -0,0 +1,59 @@
...
+output_name = sprintf('%s_ti%d', name, bin_factor);
+output_video = VideoWriter(output_name, 'MPEG-4');
+output_video.Quality = 100;
+output_video.FrameRate = behavior_video.FrameRate / bin_factor;
+open(output_video);
+
+write_idx = 0;
+for trial_idx = 1:num_trials
+    for k = 1:binned_frames_per_trial(trial_idx)
+        % Indices into the original movie
```
jmaxey: This is where I mentioned the multiplication. Its perfectly fine the way it is. Just different from what I normally do.

kimtonyhyun: I think both are fine.

kimtonyhyun:
The computation of the binned indices takes place in `bin_frame_indices2` and follows this logic:
1) For each trial, determine the number of frames in the original indices table
2) For each trial, divide the number of frames in the original by `bin_factor`, and truncate (`floor`).

This means that any frames in the original trial that do not fit into multiples of `bin_factor` will just be dropped. My approach was to drop these frames, rather than creating a binned frame with less than the number of frames (than `bin_factor`).

This means that, from each trial, `bin_factor-1` frames may be truncated -- depending on how the trial's frame count lines up with respect to `bin_factor`.

jmaxey:
Ok. That should be fine. It would be more of a problem for larger bin_factors (such as 4), however, if we are going back and extracting df/f from the full movie rather than using ic traces for event detection, it shouldn't matter.

PR #47, opened 2015-03-05
Plots path in color; blue = start, red = end

forea:
Tweaked so that user can see progression of trial, from start (blue) to end (red)

### DIFF
--- a/behavior/MouseTracking/FindTrials/mouse_path_plotter.m
+++ b/behavior/MouseTracking/FindTrials/mouse_path_plotter.m
@@ -1,7 +1,8 @@
 function [varargout] = mouse_path_plotter( movie, centroids_mat, varargin)
 % Plots the path of the mouse during each trial based on position data and
 % trial frames (either specified by plusmaze txt or extracted from movie
-% using find_start_end_of_trials)
+% using find_start_end_of_trials). Colormap = Jet --> Blue-ish = start of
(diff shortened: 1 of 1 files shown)


kimtonyhyun:
Add a picture for others to see?

forea:
Pretty colors!
screenshot from 2015-03-05 13 33 04

kimtonyhyun:
Also, I thought that if you put the trial number in each subplot not as `title` but `ylabel`, then it would be easier to distinguish which image belongs to which trial.

forea:
tried it out, hard to read

kimtonyhyun:
Ok, looks good.

behavior/MouseTracking/FindTrials/mouse_path_plotter.m
```diff
@@ -47,9 +48,12 @@
         set(gca,'XTick',[],'YTick',[]);
         colormap gray;
         hold on
-        plot(centroids(trial_start:trial_end,1),...
-             centroids(trial_start:trial_end,2),...
-             'g-');
+        
+        % plot centroids in color
+        for i = trial_start:trial_end
+            color_options = jet(size(centroids(trial_start:trial_end,1),1));
```
kimtonyhyun: Actually, this line should go outside of the `for i = 1:trial_start:trial_end` loop.

PR #48, opened 2015-03-06
reconstruct traces using the movie and the ica filters

inanhkn:
The function `reconstruct_traces` enables reconstruction of traces with custom options such as the threshold for a pixel value in an IC filter to contribute to the trace (controlled by `threshIC`), and the threshold pixel value in the `df/f` movie to contribute to the trace (controlled by `threshMov`).

example:

`reconst_traces = reconstruct_traces(movie_source,ica_filters,'threshIC',0.2,'threshMov',0)`

This reconstructs the traces ( has same array structure as in ica_traces) by discarding values in each IC filter that are lower than 0.2 times the maximum, and discarding values in the movie matrix that are lower than zero. `threshIC` must be non-negative, but `threshMov` can be set to negative values. For example, setting `threshMov` to -1 retains all the elements of the movie matrix, and setting it to -0.5 retains all the elements that are more than 0.5 times the minimum. Default for `threshMov` is 0 and the default for `threshIC` is `0.3`.

Here's an example with the defaults:

Note: The reconstructed trace was rescaled to match the scale of the IC trace.

### DIFF
--- a/image_op/reconstruct_traces.m
+++ b/image_op/reconstruct_traces.m
@@ -1,0 +1,82 @@
+function reconst_traces = reconstruct_traces(movie_source,ica_filters,varargin)
+% Reconstruct trace from the movie using the IC filters.
+%
+% inputs:
+%
+%   movie_source : link to the movie file
+%   ica_filters : ica filters corresponding to the movie
+%
+% variable input arguments:
+%
+%   threshMov : Scale for the lower threshold for the movie brightness.( Any
+%           value lower than this threshold will be set to 0). If threshMov 
+%           is negative, then the lower threshold is calculated as 
+%           |threshMov|*min(movie_matrix). If it is positive, then the lower 
+%           threshold is calculated as threshMov * max(movie_matrix).
+%           Default for threshMov is 0.
+%   threshIC : Scale for the threshold for the ICA filter weights. (any
+%           value in an ica filter lower than the threshold will be set to 
+%           0.) The threshold for an IC is calculated as 
+%           threshIC * max(ica_filter).
+%           Default for threshMov is 0.3 . 
+%   
+% output:
+%
+%   reconst_traces = [# of frames] x [# if ICs] array of reconstructed
+%   traces
+%
+% Example usage : 
+% reconst_traces = reconstruct_traces(movie_source,ica_filters,'threshIC',0.2,'threshMov',-1)
+%
+%Hakan Inan (Mar 15)
+%
+threshMov = 0;threshIC = 0.3; % Defaults
+if ~isempty(varargin)
+    len = length(varargin);
+    for k = 1:len
+        switch varargin{k}
+            case 'threshMov'
+                threshMov = varargin{k+1};
+                if ~isnumeric(threshMov)
+                    error('Movie brightness threshold must be numerical');
+                end
+            case 'threshIC'
+                threshIC = varargin{k+1};
+                if ~isnumeric(threshIC)
+                    error('IC filter threshold must be numerical');
+                end
+        end
+    end
+end
+fprintf('Loading movie.. \n');
+M = load_movie(movie_source);
+[height,width,num_frames] = size(M);
+M = reshape(M,height*width,num_frames);
+
+if threshMov<0
+    minMov = min(M(:));
+    threshMov = -threshMov*minMov;
+elseif threshMov>0
+    maxMov = max(M(:));
+    threshMov = -threshMov*maxMov;
+end
+
+num_ICs = size(ica_filters,3);
+fprintf('Thresholding IC filters.. \n');
+% Threshold the ica_filters
+for idx_cell = 1:num_ICs
+    ica_filter = ica_filters(:,:,idx_cell);
+    ica_filter(ica_filter<threshIC*max(ica_filter(:))) = 0;
+    ica_filters(:,:,idx_cell) = ica_filter/sum(ica_filter(:));
+end
+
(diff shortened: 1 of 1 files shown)


image_op/reconstruct_traces.m
```diff
@@ -0,0 +1,82 @@
...
+    minMov = min(M(:));
+    threshMov = -threshMov*minMov;
+elseif threshMov>0
+    maxMov = max(M(:));
+    threshMov = -threshMov*maxMov;
+end
+
+num_ICs = size(ica_filters,3);
+fprintf('Thresholding IC filters.. \n');
+% Threshold the ica_filters
```
kimtonyhyun: Can you use `image_op/threshold_ic_filter.m` for this operation? (You can flesh out the details of that function there.)

image_op/reconstruct_traces.m
```diff
@@ -0,0 +1,82 @@
...
+    ica_filters(:,:,idx_cell) = ica_filter/sum(ica_filter(:));
+end
+
+reconst_traces = zeros(num_frames,num_ICs);
+fprintf('Reconstructing traces.. \n');
+for idx_cell = 1:num_ICs
+    ica_filter = ica_filters(:,:,idx_cell);
+    pix_active = find(ica_filter>0);
+    movie_portion = M(pix_active,:)';
+    movie_portion(movie_portion<threshMov) = 0;
```
kimtonyhyun: Can the thresholding of the movie be performed on `M`, instead of on each `movie_portion`?

image_op/reconstruct_traces.m
```diff
@@ -0,0 +1,82 @@
...
+    reconst_traces(:,idx_cell) = movie_portion * ica_filter(pix_active);
```
kimtonyhyun: It seems like you have a very roundabout way of computing the reconstructed trace. Isn't
```
ica_mask = reshape(ica_filter(:,:,idx_cell), height*width, 1);
reconst_traces(:,idx_cell) = (ica_mask' * M)'; 
```
all that is required to reconstruct the trace?

image_op/reconstruct_traces.m
```diff
@@ -0,0 +1,82 @@
...
+%   ica_filters : ica filters corresponding to the movie
+%
+% variable input arguments:
+%
+%   threshMov : Scale for the lower threshold for the movie brightness.( Any
+%           value lower than this threshold will be set to 0). If threshMov 
+%           is negative, then the lower threshold is calculated as 
+%           |threshMov|*min(movie_matrix). If it is positive, then the lower 
+%           threshold is calculated as threshMov * max(movie_matrix).
+%           Default for threshMov is 0.
```
kimtonyhyun: Can you add note that threshMov = -1 will retain all movie pixels?

image_op/reconstruct_traces.m
```diff
@@ -0,0 +1,82 @@
...
+[height,width,num_frames] = size(M);
+M = reshape(M,height*width,num_frames);
+
+if threshMov<0
+    minMov = min(M(:));
+    threshMov = -threshMov*minMov;
+elseif threshMov>0
+    maxMov = max(M(:));
+    threshMov = -threshMov*maxMov;
+end
```
kimtonyhyun: The behavior of `threshMov` is kind of complicated. Why not let `threshMov` range between 0 and 1, and you can do:
```
minMov = min(M(:));
maxMov = max(M(:));
threshMov = (maxMov-minMov)*threshMov + minMov;
```
i.e. `threshMov == 0` keeps all pixels, and `threshMov == 1` gets rid of everything?

kimtonyhyun:
Left some minor suggestions. Can you briefly explain on why we need `threshMov`?

inanhkn:
I'll use threshold_ic_filter for thresholding the ica filters. If you can tell me a way to threshold the movie without replicating it, I can do the matrix multiplication without looping through the ICs. I tried M(M<threshMov) ==0 and it somehow still doubled the memory usage, or maybe I hallucinated :)

Also by multiplying with `movie_portion` instead of `M`, there is a huge saving from the number of rows of the movie matrix. I don't know how the trade off is between this and movie slicing+find, I kind of verified it is not bad. Specifically I did something equivalent to:
`ica_mask = reshape(ica_filters, height*width, num_ICs);
reconst_traces_v2 = (ica_mask' * M)'; `
 This is what I meant earlier by not having to loop. This took more or less the same amount of time as what the script is currently using. This is how I can justify constructing `movie_portion` (also the problem of thresholding the whole movie makes this the most viable option).

I want to keep the threshMov the way it is since scaling it between -1 and 1 enables to consciously threshold below/above 0 point. You might argue 0 point might not be very informative in a df/f movie, but I still think it is good practice to scale between -1 and 1. I'll state it in the comment that -1 leaves the movie the way it is.

inanhkn:
I like threshMov by the way :) One can tweak the threshold by an arbitrary amount to achieve a trade off between being close to the ica trace and being less noisy.

kimtonyhyun:
Ok, thanks for the explanation. Looks good to me after refactoring to use `image_op/threshold_ic_filter`.

If the reconstructed traces generally look good, then we can also update `run_ica` to run `reconstruct_traces` so that each `ica_(...).mat` file also has `reconst_traces` in addition to `ica_traces`.

kimtonyhyun:
How about calling these types of traces (masked IC filters applied to the DFF movie) "reextracted" traces, rather than "reconstructed" traces? Thoughts?

inanhkn:
Whatever floats your boat man :)

PR #50, opened 2015-03-06
File-to-file cropping

kimtonyhyun:
I am now working on implementing our preprocessing steps (crop, motion-correction, norm-by-disk, DFF) so that each step works from "file-to-file".

This PR contains the crop operation performed file-to-file, in `crop_movie`. It loads chunks of the movie (2500 frames at a time, which is a parameter), crops the movie chunk, and saves it to an output file.

I also added a `readme.md`, as an initial pass at the _specification_ for our HDF5 files. In particular:
- `/Data/Images` is the location of the movie in the HDF5 file
- `/Params/*` contains parameters that all HDF5 movies are _expected_ to have. As the preprocessing functions generate new HDF5 files, it is responsible for copying over the `/Params/*` datasets to the new file and updating if necessary. The helper function `copy_hdf5_params` takes care of this transfer.

I also edited the original `concatenateHDF5` to not write `NumFrames`, `NumRows`, `NumCols` into `Params`. These movie dimensions can be obtained directly from `/Data/Images` (without having to read the entire movie).

The updated `concatenateHDF5` and `crop_movie` have been verified to work with c9m7d12 data.


inanhkn:
looks good to me.

PR #51, opened 2015-03-06
break down session into trials

inanhkn:
`summarize_trials` makes a summary of each session (summary elements are traces, start arm,finish arm and trial correctness). The output is a .mat file containing a [# of trials] x 4 cell array.

Example: 
`summarize_trials(sources,ica_traces,'includeProbes',0,'classFile','class_dummy.txt')`
above call leaves the probe trials out and also only considers the ICs that are classified to be cells in `'class_dummy.txt'` to generate its output.

### DIFF
--- a/io/summarize_trials.m
+++ b/io/summarize_trials.m
@@ -1,0 +1,69 @@
+function summarize_trials(sources,traces,varargin)
+% Outputs a .mat file with [# of trials] x 4 cell having the following 4 columns:
+%   trial start arm
+%   trial finish arm
+%   trial correctness
+%   cell traces belonging to the trial
+%
+% inputs:
+%   sources : Struct containing maze output, trim and binning parameters
+%   needed here.
+%   traces : [# of frames] x [#of ICs] array of cell traces.
+%
+% variable input arguments:
+%   includeProbes: Set to 0 to retain all the trials, set to 1 to get rid
+%   of probe trials.
+%   classFile: Path to the classification file. Specifying this will
+%   generate output for only the ICs that are classified as cells.
+%
+% Hakan Inan (Mar 15)
+%
+include_probes = 1; % Default
+if ~isempty(varargin)
+    len = length(varargin);
+    for k = 1:len
+        switch varargin{k}
+            case 'includeProbes'
+                include_probes = varargin{k+1};
+                if include_probes~=0 && include_probes~=1
+                    error('includeProbes can be either 0 or 1.');
+                end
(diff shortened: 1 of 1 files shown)


kimtonyhyun:
Is the idea that you'll expand this in the future to take into account behavior centroid positions, etc.?

inanhkn:
Yes, that is a natural expansion of this.

kimtonyhyun:
By the way, `concatenateHDF5` already maps N/S/E/W onto numbers. If we ever do the numeric mapping elsewhere, we should make it consistent.

https://github.com/inanhkn/analysis/blob/master/io/concatenateHDF5.m#L145

jmaxey:
These were just values that I randomly assigned. If someone else has a different mapping in mind, feel free to change it!

----- Original Message -----

kimtonyhyun:
What this `summary` (of a day) currently lacks is the ICA filters. I think that the `summary` would be more generally useful if the returned value was a nested struct with a form like:
```
summary.cells = [height x width x num_ICs]
summary.trials = (... cell structure that you have now ...)
```
(Note, we could also then expand the `cells` subfield later, to include things like boundaries, masks, i.e. other properties of the cells of that day, but ones that apply to all trials.)

Also, do you prefer to use a [num_trials x 4] cell to represent the start/goal/actual arms of the trial? An alternative to consider would be to use a struct representation of each trial (with fields `start`, `goal`, `actual`), so that instead of numeric indices (1-3), you could write things like: `summary.trials(5).start` which may be more easily readable.

kimtonyhyun:
In fact, I would be in favor of promoting `summary` as returned by this function to an object. This would have the following advantages:
1) Can easily do type testing in subsequent functions -- i.e. a function such as `view_superimposed_trials` can test its input arguments to see if it is a `DaySummary` datatype (via [`isa`](http://www.mathworks.com/help/matlab/ref/isa.html) function). We then would have standardized visualization functions that work with the `DaySummary` datatype.
2) Once we shove in ICA filters, masks, etc. the memory footprint of `summary` will increase (not anywhere near the actual DFF movie though). By writing as a class, the memory contents will not be replicated as the `summary` is provided into functions.

kimtonyhyun:
Basically, I was also inspecting ICs across days, and found myself needing ICA images, traces, trial frame indices, start arms, correctness, etc... and agreed that I would benefit from more organization such as a `summary`.

kimtonyhyun:
Closing in favor of #68

PR #52, opened 2015-03-07
constrain the mouse spatial trace to a line and reparametrize

inanhkn:
@forea 
`reparameterize_centroids` takes in the centroids of the mouse in a trial as a  [# of frames in trial] x 2 array and outputs 2 things:
`intermediate_centroids` : This is the output of fitting a segmented line to the centroids. You can use this to have a smooth spatial trace overlay on the actual centroids.
`param_centroids` : output of translating and rotating `intermediate_centroids` so that the centroids vary only in vertical and horizontal directions ([0,0] is the stem).

Note: there is a commented out portion at the end of the script that applies median filtering to the final centroids. I don't remember why I used to use it, but feel free to uncomment it and see if it becomes any better. I haven't tried the code in a long time, so I would appreciate if you could try as many trials as possible and tell me how it performs.

### DIFF
--- a/behavior/MouseTracking/reparameterize_centroids.m
+++ b/behavior/MouseTracking/reparameterize_centroids.m
@@ -1,0 +1,89 @@
+function [intermediate_centroids,param_centroids] = reparameterize_centroids(centroids)
+    % Re-parametrize the spatial trace such that:
+    %   south is (-*,0)
+    %   north is (+*,0)
+    %   east is (0,+*)
+    %   west is (0,-*)
+    %
+    %   spatial_trace : [# of frames] x 2 array
+    % 
+    %   param_trace : parametrized spatial trace with the same size as
+    %   spatial_trace
+    %
+    % Hakan Inan (Mar 15) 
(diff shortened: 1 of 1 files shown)


behavior/MouseTracking/reparameterize_centroids.m
```diff
@@ -0,0 +1,89 @@
...
+    b = -dum_prime.*coord_x+dum.*coord_y;
+    u = A\b;
+    slope = u(1);
+    intercept1 = u(2);
+    intercept2 = u(3)/-slope;
+    
+    % Reconstruct y trace from x trace
+    y_reconst = dum.*(slope*coord_x+intercept1)+...
+        dum_prime.*(-1/slope *coord_x + intercept2);    
+    intermediate_centroids = [coord_x,y_reconstruct];
```
forea: is this supposed to be
intermediate_centroids = [coord_x,y_reconst] ?

inanhkn:
corrected it.

PR #54, opened 2015-03-07
Move Miniscope-specific files into subdir

kimtonyhyun:
I'd like to start organizing the `io` directory, which is getting bloated as of late.

@forea This PR moves files that are very specific to the Miniscope into its own subdirectory `io/miniscope`. These are the files that we use, for instance, to make sure that we copied all the data from the recording computers, and that the (Inscopix) decompressor worked as expected.


kimtonyhyun:
@jmaxey Please note that you'll now have to add `io/miniscope` directory to your Matlab path for `concatenateHDF5`. (It needs the `parse_miniscope_xml`.)

PR #55, opened 2015-03-07
Routines for file-to-file movie preprocessing. 

kimtonyhyun:
Currently, the preprocessing steps (starting from the concatenated movie) are as follows:

1. Crop movie, to get rid of not-useful areas in the field-of-view
2. Motion correction,
3. Crop movie, to get rid of "black borders" introduced by motion correction
4. Norm movie (by disk filter)
5. Compute DFF
6. Temporally bin DFF (for PCA/ICA)

The individual steps are implemented through `crop_movie`, `register_movie`, `norm_movie`, which take in an input HDF5 file, and write the result to an output HDF5 file. As described in `io/readme.md`, each function will carry over the contents of `/Params/*` datasets from file-to-file.

I also introduced a helper function `get_dataset_info(hdf5_source, dataset_name)` which retrieves the dimensions of the dataset in the HDF5, as well as its type. It is used to determine the size of the movie matrix in the HDF5 file, and its type (i.e. `uint16` vs. `single`).

These files were successfully used to complete the preprocessing steps for c9m7d12 -- from the original concatenated movie to DFF (both full frame and 2x time-binned).

Also, minor tweaks to `compute_movie_scale` and `view_movie` which I found helpful for working with both raw, unnormalized movies (in addition to the DFFs that we usually look at).


kimtonyhyun:
Note that temporal binning must occur at the very end of the preprocessing chain (after DFF at full frame rate), if we want the option of re-extracting full frame rate traces via the ICA masks.

I'm updating `bin_movie_in_time` to work as file-to-file right now.

kimtonyhyun:
`bin_movie_in_time` has also been update to work from file-to-file.

kimtonyhyun:
With these files, I should be able to plow through Mouse 7, Days 6 through 12 over the next few days.

kimtonyhyun:
Assigned to @jmaxey since she's going to run the file-to-file scripts for Day 26.

jmaxey:
The new register movie should have a prompt instructing the user to select the ROI. Right now nothing comes up and it just sit. It took me a couple of seconds to realize what I needed to do.

kimtonyhyun:
Added a command line prompt asking for ROI selection in `register_movie`.

jmaxey:
Would it be possible to change these so that the automated output file (assuming the user inputs '[]') is saved to the same directory that the input file is coming from? I keep running the files and having the output saved to the '/analysis/image_op' directory.

kimtonyhyun:
Ok, I'll do that.

jmaxey:
Just an update: I am having trouble running the register_movie function. It crashes on the first h5create command every time. Have you seen this before? It's possibly a local thing that I need to work out with my computer.

kimtonyhyun:
Hm... I haven't seen that before. Maybe it's a permissions thing -- I
always run the script from where I want the output file and where I have
write access. We'll come over to CNC for lunch, and maybe we can look at it
then,

-Tony

2015-03-24 9:36 GMT-07:00 jmaxey <notifications@github.com>:

jmaxey:
Sounds good. I don't think its a permissions thing though. I have tried running it both in the directory I want the file to be saved and by specifying the directory in the output name and get the same error. And I've tried to save it both the desktop and the external hard drive.

----- Original Message -----

kimtonyhyun:
Are you actually putting in `'[]'` (string) for the output filename, as you noted in an earlier comment? The script handles `''` for default filename (the empty string) -- but not `'[]'`. I'd imagine `[]` (the empty array) will also work correctly.

jmaxey:
When you crop the movie after motion correction, which function do you use? crop_movie.m automatically brings up frame 1 but frame 1 will not have black around the edges as the other frames were registered to it. Can ref_idx be an option for varagin?

kimtonyhyun:
I tend to do the following:

```
M = load_movie('movie_to_be_cropped.hdf5');
F = compute_fluorescence_stats(M, 'trim', 8);
plot(F);
```

Each row of `F` contains the `[min mean max]` fluorescence value of each frame. The optional arguments `'trim', 8` indicates to trim 8 pixels from the border (all four sides) in the fluorescence stat computation.

Once I observe that the minimum fluorescence value of each frame is greater than 0, then I run:
```
crop_movie('movie_to_be_cropped.hdf5', '', 'trim', 8);
```
where the optional arguments `'trim', 8` has the same interpretation as `compute_fluorescence_stats`.

jmaxey:
Is there a way to do this without reading the entire movie into memory?

kimtonyhyun:
Yes, `F = compute_fluorescence_stats('movie_to_be_cropped.hdf5', 'trim', 8);` will do the same thing, but reading from file.

jmaxey:
Perfect. Thanks

jmaxey:
Ok. I've run everything on my computer and it seems to work. The only thing that I think should be changed is saving the output file to the same directory as the input.

kimtonyhyun:
@jmaxey I promise to follow up with a PR that makes the output file show up at the same directory as the input. I'm merging for now -- have another branch that needs changes from this one.

PR #56, opened 2015-03-08
Move tiff functions

kimtonyhyun:
This moves TIFF related functions into their own subfolder: `io/tif`.

While we won't be using TIFF anymore for concatenating the Miniscope movies, the loaders are still useful to have around (for instance, when we need to inspect the original TIFF files from the Miniscope).


jmaxey:
The TIFF functions actually are still used in concatenateHDF5 so will also need to be added to the path.

kimtonyhyun:
That's cool with me

PR #57, opened 2015-03-08
Update classify ICs

kimtonyhyun:
#### Do not merge this PR until everyone has agreed!

This PR contains several fixes and improvements pertinent to `classify_ics` and `run_alignment` (cross-day alignment tool) -- to improve the computation and display of IC boundaries. In particular, one needs to use this branch for running cross-day alignment.

#### First, `classify_ics`:

_Significant change to text-file behavior_: Previously, we would read the frame indices from the raw text file, then perform the operations in `classify_ics` that would make it compatible with the trimmed and binned movie (e.g. `compress_frame_indices` and `bin_frame_indices`). Instead, `classify_ics` now expects that the text-file indicated in `data_sources.maze` contains the indices that are compatible with the movie without any further processing of the indices. (Later, we will pull these indices directly from the HDF5 file.)

The new `data_sources` structs should look something like:
```
sources.maze = 'c9m7d25_ti2.txt'; % Contains indices that have been trimmed/binned!
sources.miniscope = 'c9m7d25_dff_ti2.hdf5';
sources.fps = 10;
sources.pca = 'pca_n500.mat';
```

_Support for multiple boundaries during classification_: The current behavior (in `master` branch) of `classify_ics` is that only one boundary will be drawn for each IC, even if the underlying IC filter is distributed. This PR makes it so that all boundaries corresponding to a certain `ic_filter_thresh` will be drawn, thereby yielding plots like:
multiple-boundaries2
multiple-boundaries

This makes it much easier to see if the IC filter is not localized in the field of view.

_Draw the center of mass with a blue dot_:

I've noticed that the boundary computed from the filter can sometimes give a mis-representation of the underlying IC filter. This is so, because the IC filter is weighted, and could be giving some pixel a whole lot more weight than others.

In order to visualize this issue, `classify_ics` now draws a blue dot at the center of mass of the IC.

A good IC filter will then look like this:
com-dot

On the other hand, a bad IC filter may look like this (it may be being influenced by hot pixels):
com-dot_bad

As a corollary, I now think that we should be re-extracting the traces from the movies, using a modified form of the IC filter. In particular, we should enforce filter locality. (The above pictures come from c9m7d25.)

_Loading previously-stored classification file during `classify_ics`_: A minor convenience update. When you load in a previously-stored classification file, the script jumps to the first IC that has not been classified.

#### Second, `run_alignment`:

When manually selecting the control points (ICs) for cross-day alignment in `run_alignment` (specifically, in `compute_affine_transform`), there was a bug that would sometimes cause the script to crash.

Conceptually, the root cause of the crash was that there can be significant discrepancy between the displayed boundaries and the underlying IC mask. I've resolved this discrepancy by making the control point selection rely solely on the displayed boundaries, and _not_ the underlying mask.


inanhkn:
I agree to the changes. As a step further, how about modifying trace reconstruction procedure to only consider the most active blob when an IC filter has multiple blobs retained even above a certain activity threshold?

kimtonyhyun:
That's an interesting idea. In the case of multiple blobs per IC, I was going to manually select the blob to keep (probably based on size). On the other hand, we could split up the blobs and score them individually as you say.

One thing that's emerging is that it's not just "reconstructed traces" but "reconstructed traces _and_ filters."

inanhkn:
Yes, I like it :)
By reconstructing IC filters, we finish what ICA started and left incomplete.

forea:
I agree to the changes.

In the case of multiple ICs per blob, I am in favor of splitting them up and evaluating them individually rather than only taking the "best" one based on some metric (size, activity) -- important for cases where the multiple blobs are all good cells that happen to be highly correlated (I remember day 7 or 8 has a pair like that).

forea:
RE: As a corollary, I now think that we should be re-extracting the traces from the movies, using a modified form of the IC filter.

>> what do you mean by modified form of IC filter?

kimtonyhyun:
The idea is that some IC filters are not very local -- the filter extends significantly more in the field-of-view than is reasonable for a cell. In such cases, it may be good to "modify" or process the IC filter before using it as a mask to resample the trace from the movie. For example, we may truncate the spatial extent of the IC, thereby yielding an ROI that is very spatially localized around a cell.

jmaxey:
I agree with Fori on how to deal with multiple blobs per IC. 

For the manually processed filter, would the user need to draw the ROI boundary for the new IC?

kimtonyhyun:
Yeah, we can write the ICA filter --> ROIs routine in such a way that it supports splitting of IC filters, etc. I've also seen IC cells that encompassed two separate cells.

I don't think the user need to do the ROI boundary _manually_. I think `compute_ic_boundary` does a decent job finding boundaries.

inanhkn:
Here's my comment on splitting IC blobs and processing separately:
What PCA does is it gives combination of many cells in each PC filter. It is ICA's job to separate these mutliple cells in a PC filter. However, Most times it will not perfectly cancel out other cells in a PC filter since it is purely linear and it needs a lot of good measurements(PC filters) to linearly subtract all other cells in every PC filter. 
Given this background info, I think ICA would want us to eliminate every blob that has lower activity than the highest, because the other blobs are probably already the most active blob in another IC filter. In most of the cases, ICA could not perfectly eliminated other blobs when extracting one. 
You're right, it might make sense to examine each blob separately, but making sure that that same blob does not exist in any other IC filter.

kimtonyhyun:
@jmaxey Are you good with the changes to `classify_ics` as described in this PR?

jmaxey:
They sound good to me.

PR #58, opened 2015-03-09
Fori/check centroids

forea:
This function plays the behavior movie for specified trials with centroid plotted on top.

Features:
- Pause in between each trial (black frame)
- Pause on the first frame of each trial, with the first frame's centroid plotted in magenta as opposed to green. This is useful for checking bleed-through across frames (make sure the first centroid of the trial is not at the mouse's position from previous trial's end). Note in future version of mouse tracking this should not be an issue.

Example screenshot:
screenshot from 2015-03-09 13 37 24

### DIFF
--- a/behavior/MouseTracking/watch_trial_with_centroids.m
+++ b/behavior/MouseTracking/watch_trial_with_centroids.m
@@ -39,7 +39,7 @@ function watch_trial_with_centroids( movie, centroids_mat, trial, varargin )
     trial_start = this_trial_indices(1);  
     first_image = read(behavior_vid,trial_start);
     k = imagesc(first_image);
-    title('Trial');
+    title(sprintf('Trial %d',trial(1)),'FontSize',18);
     axis image; colormap gray;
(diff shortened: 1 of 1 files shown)


behavior/MouseTracking/watch_trial_with_centroids.m
```diff
@@ -0,0 +1,87 @@
...
+%
+% 2015-03-06 Fori Wang
+%
+%
+
+    % get trial indices
+    if ~isempty(varargin)
+        trial_indices = varargin{1};
+        trial_indices = get_trial_frame_indices(trial_indices);
+        trial_indices = [trial_indices(:,1) trial_indices(:,4)];
```
kimtonyhyun: Just fyi, `trial_indices = trial_indices(:,[1 4])` does the same.

kimtonyhyun:
That is an interesting screenshot -- looks like the Miniscope cable is disappearing into thin air.

PR #59, opened 2015-03-10
Fori/trial aware tracking

forea:
If user provides trial text file, in the cases when mouse cannot be found (sometimes happens if wire crosses over the mouse, making mouse blob small), if the frame is at the start of a trial, function uses next centroid as proxy as opposed to previous centroid to avoid trial bleed-through.

### DIFF
--- a/behavior/MouseTracking/get_mouse_XY_pos.m
+++ b/behavior/MouseTracking/get_mouse_XY_pos.m
@@ -2,6 +2,14 @@ function [ centroids ] = get_mouse_XY_pos( movie, varargin )
 % Extract the X,Y positions ( centroids ) of the mouse from the behavior
 %   video ( movie ), with live monitor option (specify 'displayTracking')
 %  
+% Example uses:
+%     Normal mode:
+%       centroids = get_mouse_XY_pos('c9m7d08_ti2-sub.mp4');
+%   Live monitor mode:	
+%         centroids = get_mouse_XY_pos('c9m7d08_ti2-sub.mp4','displayTracking');
+%     Trial-aware mode:
+%       centroids = get_mouse_XY_pos('c9m7d08_ti2-sub.mp4','c9m7d08_cr_ti2.txt');
+% 
 % Input:
 %   - movie: Behavior video (m4v); movie should be cropped (wavy curtain
 %   leads to fall centroids (haven't found another way to fix this)
@@ -19,11 +27,8 @@ function [ centroids ] = get_mouse_XY_pos( movie, varargin )
 %   number of frames in movie
 %
 % Notes: Analyzes movie in chunks of chunk_size=1000 frames (smaller memory load)
-%   *** read function does not work properly on Linux, so unfortunately
-%   have to run this on a Windows machine ***
-%   *** Not very fast, but only have to run once; ~54K frames = ~40 min
 %
-% 2015-02-26 Fori Wang
+% Updated 2015-03-10 Fori Wang
 
     display_tracking = 0;
     % check if live monitor option requested
@@ -34,8 +39,6 @@ function [ centroids ] = get_mouse_XY_pos( movie, varargin )
             figure;
             fprintf('Displaying live tracking...');            
         else
-%             fprintf('Did not recognize %s.',option);
-%             return
             trial_aware = 1;
             trial_indices = get_trial_frame_indices(option);
             trial_indices = trial_indices(:,[1 4]);
@@ -63,9 +66,6 @@ function [ centroids ] = get_mouse_XY_pos( movie, varargin )
     bg_image = uint8(bg_image);
     
     c_old = [0 0];
-    
-
-    
     
     for idx = 1:length(frame_chunks)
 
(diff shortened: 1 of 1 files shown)


behavior/MouseTracking/get_mouse_XY_pos.m
```diff
@@ -2,24 +2,33 @@
...
 % Input:
 %   - movie: Behavior video (m4v); movie should be cropped (wavy curtain
 %   leads to fall centroids (haven't found another way to fix this)
-%   - varargin: 'displayTracking' displays side-by-side of original movie and
-%   processing movie, along with centroid on each. blue * means good
-%   centroid value, red * means could not find mouse, using previous
-%   centroid value
+%   - varargin:
+%     'displayTracking': Displays side-by-side of original movie and
+%           processing movie, along with centroid on each. blue * means good
```
kimtonyhyun: Is the difference between "original" and "processing" movie that the latter is, for example, a binarized, masked version of the original frames?

behavior/MouseTracking/get_mouse_XY_pos.m
```diff
@@ -28,10 +37,11 @@
...
             display_tracking = 1;
             figure;
-            fprintf('Displaying live tracking...');
+            fprintf('Displaying live tracking...');            
         else
-            fprintf('Did not recognize %s.',option);
-            return
+            trial_aware = 1;
+            trial_indices = get_trial_frame_indices(option);
+            trial_indices = trial_indices(:,[1 4]);
```
kimtonyhyun: FYI, the downside of structuring `varargin` processing this way is that you can't do "live tracking" if you are providing the trial indices text file.

kimtonyhyun:
What happens with this code if the mouse can't be detected for the first two frames of a trial?

behavior/MouseTracking/get_mouse_XY_pos.m
forea: yes. Should I clarify that in the comments?

behavior/MouseTracking/get_mouse_XY_pos.m
forea: Yeah, I know. I haven't yet looked into how to really use varargin to the fullest.

behavior/MouseTracking/get_mouse_XY_pos.m
kimtonyhyun: Nah, it just took me a few seconds to realize what you had meant here (and why a "side-by-side" is needed when only a single movie is going into the function).

behavior/MouseTracking/get_mouse_XY_pos.m
kimtonyhyun: Take a look at Hakan's code for `varargin` here: https://github.com/inanhkn/analysis/blob/master/image_op/reconstruct_traces.m#L35

forea:
I have a counter called lost_centroids. The function keeps going until it finds the mouse, and then all previous frames are assigned to this position.

see line 151:
centroids(true_frame_idx-lost_centroids:true_frame_idx-1,:) = c_new

kimtonyhyun:
I assumed that was the role of `lost_centroids`. However, it's not clear to me that this counter is incremented appropriately for frames that are not the first of a trial.

Specifically, `lost_centroids` is incremented only when `~isempty(find(true_frame_idx==trial_indices(:,1),1))` is true. This can only occur for the first frame of a trial, as far as I can tell.

behavior/MouseTracking/get_mouse_XY_pos.m
```diff
@@ -112,14 +126,33 @@
...
             % Save centroids data and update plot
+            true_frame_idx = frame_chunks(idx,1)+frame_idx-1;           
+                
             if isempty(area_vector) %sometimes blob disappears, go to previous centroid
-                centroids(frame_chunks(idx,1)+frame_idx-1,:) = c_old;
-                centroid_color = 'r';
+                
+                % trial_aware mode
+                % if image is the first image of a trial, use next centroid
+                if trial_aware && ~isempty(find(true_frame_idx==trial_indices(:,1),1))
```
kimtonyhyun: This can be written simply (but double check):
```
if trial_aware && find(true_frame_idx==trial_indices(:,1),1)
```
since `find` will return an empty list (`[]`) when there is no match; and the empty list is equivalent to a false in Matlab.

forea:
> Specifically, lost_centroids is incremented only when ~isempty(find(true_frame_idx==trial_indices(:,1),1)) is true. This can only occur for the first frame of a trial, as far as I can tell.

You are right. I'll fix that.

kimtonyhyun:
Also, do we have the "behavior dropped table" for Day 10? I think I'd like to make another side-by-side for Day 10.

forea:
Yep, here it is: https://www.dropbox.com/s/a6aklkpwwl73kf3/mouse7_day10_allo-south_dropped_behavior_frames.txt?dl=0

behavior/MouseTracking/get_mouse_XY_pos.m
```diff
@@ -112,14 +126,38 @@
...
-                centroids(frame_chunks(idx,1)+frame_idx-1,:) = c_old;
-                centroid_color = 'r';
+                
+                % trial_aware mode
+                % if image is the first image of a trial, use next centroid
+                if trial_aware && find(true_frame_idx==trial_indices(:,1),1)
+                    reassign_prev_centroid = 1;
+                    lost_centroids = lost_centroids+1;
+                    centroid_color = 'm';
+                elseif trial_aware && reassign_prev_centroid == 1
```
forea: added this in for cases when mouse is still not found in trials following first frame

kimtonyhyun: This can just be `elseif reassign_prev_centroid`

forea:
Updated

PR #64, opened 2015-03-12
Updated to new plusmaze data format

forea:


behavior/calc_behavior_score.m
```diff
@@ -1,16 +1,12 @@
...
 
-    fid = fopen(file);
-    maze_data = textscan(fid,'%s %s %s %f %d %d');
-    fclose(fid);
-
-    end_arm = maze_data{2};
-    actual_end_arm = maze_data{3};
-    num_trials = length(actual_end_arm);
+    [~,maze_data,~] = parse_plusmaze(file);
+    end_arm = maze_data(:,2);
```
kimtonyhyun: I think `goal_arm` would be clearer, but this works too.

PR #68, opened 2015-03-16
Kimth/day summary

kimtonyhyun:
@inanhkn This is a re-implementation of PR #51 where the "summary" is implemented as an object.

The summary object can be instantiated as follows:
```
>> ds = DaySummary('_data/c9m7d06_ti2.txt', 'ica001', 'excludeprobe')

ds = 

  DaySummary with properties:

            cells: [250x1 struct]
           trials: [100x1 struct]
        num_cells: 250
       num_trials: 100
    trial_indices: [100x4 int32]
```
The first argument to the constructor is the PlusMaze text file, where the frame indices contained in the text file should be consistent with the ICA traces. The second argument to the constructor is the ICA directory, which should contain the ICA result (`ica_*.mat`) and the classification result (`class_*.txt`).

In the near future, we should augment `DaySummary` to take in also the mouse positions within each trial.

Once the object has been returned (as `ds`, in this example). You can check if the variable is of a `DaySummary` type:
```
>> isa(ds, 'DaySummary')

ans =

     1
```
This type-checking ability should allow us to build visualization / analysis functions that take in `DaySummary` objects.

You can also access trial information as follows:
```
>> ds.trials(4)

ans = 

     start: 'east'
      goal: 'south'
       end: 'south'
      time: 24.6150
    traces: [250x236 single]

>> ds.trials(4).start

ans =

east
```

You can access the trace of trial `k` and IC index `m` as follows: `ds.trials(k).traces(m,:)`, which is similar to your access pattern in #51. However, for convenience I also added a convenience method that allows you to retrieve the same trace by: `ds.get_cell_trace(m, k)`.

An aside: one key difference between an object and a struct is that the former allows you to associate functions (like `get_cell_trace`) -- called "methods" -- with the object. This is in contrast to structs which does not have that capability.

### DIFF
--- a/io/DaySummary.m
+++ b/io/DaySummary.m
@@ -1,0 +1,81 @@
+classdef DaySummary
+% Summary of PlusMaze data for a single day.
+%
+% Inputs:
+%   plusmaze_txt: Name of the PlusMaze output text file. Frame indices of
+%       the text file needs to be consistent with the ICA traces!
+%   ica_dir: Directory containing 
+%       - ICA results in a "ica_*.mat" file
+%       - Classification results in a "class_*.txt" file
+%
+% Output:
+%   DaySummary object
+%
+% Example usage:
+%   ds = DaySummary('c9m7d06_ti2.txt', 'ica001');
+%
+    properties
+        cells
+        trials
+        
+        num_cells
+        num_trials
+        
+        trial_indices
+    end
+    
+    methods
+        function obj = DaySummary(plusmaze_txt, ica_dir, varargin)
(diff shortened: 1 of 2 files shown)


kimtonyhyun:
@inanhkn After you take a look at this `DaySummary`, can you modify your multi-day plotting function (like the one you showed us this morning) to work with `DaySummary`? I'd also like to browse through some of the ICs.

inanhkn:
It looks great, thanks. 
I guess you will write your own multi-day plotting function now :)

PR #69, opened 2015-03-17
Kimth/multiday plots

kimtonyhyun:
@inanhkn A simple basic built-in visualization function in `DaySummary`, called `plot_cell_phase`.

Example usage:
```
ds = 

  DaySummary with properties:

            cells: [250x1 struct]
           trials: [110x1 struct]
        num_cells: 250
       num_trials: 110
    trial_indices: [110x4 int32]

>> ds.plot_cell_phase(4);
```
will plot the "superimposed trial plot" for IC 4 of that `DaySummary` as follows:
daysummary_plotcellphase

### DIFF
--- a/io/DaySummary.m
+++ b/io/DaySummary.m
@@ -99,6 +99,8 @@ classdef DaySummary
                 vararg = varargin{k};
                 if ischar(vararg)
                     switch lower(vararg)
+                        case 'incorrect'
+                            display_trial = ~strcmp({obj.trials.goal}, {obj.trials.end});
                         case 'correct'
                             display_trial = strcmp({obj.trials.goal}, {obj.trials.end});
                         case 'start'
--- a/io/DaySummary.m
+++ b/io/DaySummary.m
@@ -101,6 +101,36 @@ classdef DaySummary
         % Built-in visualization functions
         % Note: Do NOT make use of subplots in the built-in plot methods
         %------------------------------------------------------------
+        function plot_trace(obj, cell_idx)
+            % Plot the trace of a single cell, color-coded by trial
+            trace_min = Inf;
+            trace_max = -Inf;
+            
+            colors = 'kbr';
+            for k = 1:obj.num_trials
+                [trace, inds] = obj.get_trace(cell_idx, k);
+                
+                trial_trace_min = min(trace);
+                trial_trace_max = max(trace);
+                if (trial_trace_min < trace_min)
+                    trace_min = trial_trace_min;
+                end
+                if (trial_trace_max > trace_max)
+                    trace_max = trial_trace_max;
+                end
+                
+                plot(inds, trace,...
+                     colors(mod(k,length(colors))+1));
+                hold on;
+            end
+            hold off;
+            xlim([1 inds(end)]); % Using the last 'inds' from loop!
+            trace_delta = trace_max - trace_min;
+            ylim([trace_min trace_max] + 0.1*trace_delta*[-1 1]);
+            xlabel('Frame index');
+            ylabel('Signal [a.u.]');
+        end
+        
         function plot_superposed_trials(obj, cell_idx, varargin)
             % Optional arguments allow for filtering of trials, e.g.
             %   "plot_superposed_trials(cell_idx, 'start', 'east')"
--- a/io/DaySummary.m
+++ b/io/DaySummary.m
@@ -1,4 +1,3 @@
-classdef DaySummary
 % Summary of PlusMaze data for a single day.
 %
 % Inputs:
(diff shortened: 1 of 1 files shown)


inanhkn:
I think it would be nice if we could plot particular trials of a cell, .e.g trials in [1:20] or trials starting on the east arm. I guess the second example can be manually done if we have the option to plot only the given trials.

kimtonyhyun:
Updated to allow for trial filtering. If `ds` is a `DaySummary` object, then
```
ds.plot_superposed_trials(k, 'start', 'east');
```
will plot only the trials of cell `k` if they start from the east arm. Right now, there are filters for start and end arms. More can be added in the `varargin` loop.

kimtonyhyun:
Here are examples of the multiple visualization functions built-in into `DaySummary`:

```
>> ds = DaySummary('_data/c9m7d06_ti2.txt', 'ica001');
>> ds.plot_trace(6);
```
results in the trial color-coded trace of IC 6:
plot_trace

Or,
```
>> ds.plot_superposed_trials(6);
```
results in the trial superposition of IC 6:
plot_superposed_trials

One more,
```
>> ds.plot_cell_raster(6);
```
gives:
plot_cell_raster

kimtonyhyun:
I am finding that the raster plot is really helpful, in addition to the other two plots that we've been using a lot until this point.

Here's a combined view (not built-in in `DaySummary`):
combined

kimtonyhyun:
What's really nice about the raster (I think) is that it hides away the fact that the ICA trace may not be beautiful, but that there is a very strong trial-repeatability:

combined2

forea:
this is nice!

jmaxey:
These raster plots look nice. Do you want me to PR my version? It seems like this version does the same thing and is already incorporated into the function.

kimtonyhyun:
@jmaxey You should try out `DaySummary` to see if there are missing functionalities, etc. I found that having such an encapsulated object for 1 day is basically necessary for sanity, as I look at multiple days.

jmaxey:
For a single cell it looks good. I'll keep cleaning the function that looks at all cells across a single trial (which does't seem to be included). I'll also submit some example images.

kimtonyhyun:
Yeah, I thought that would be good (all cells in a single trial, preferably sorted by firing onset).

jmaxey:
Are the trial indices stored in the structure already binned according to the downsampling of the movie?

kimtonyhyun:
Yes, initialization of `DaySummary` requires that the text file input contains indices that are already consistent with the downsampling (i.e. the temporal length of ICA traces).

If you are at the Herrin computer, you can see that all the DFF movies I've generated so far for Mouse 7 (look at drive E, I think), have `c9m7dXX.txt` and `c9m7dXX_ti2.txt`. The former is the "trimmed" frame indices compatible with the concatenated movie (`concatenateHDF5`). The latter is the binned indices compatible with the binned movie.

forea:
got an error.. what am I missing?

>> ds = DaySummary('_data7/mouse7_day07_ego-left_ti2.txt', 'd7_ica001',
'excludeprobe')

ds =

  DaySummary with properties:

            cells: [250x1 struct]
           trials: [100x1 struct]
        num_cells: 250
       num_trials: 100
    trial_indices: [100x4 int32]

>> ds.plot_cell_phase(4);
No appropriate method, property, or field plot_cell_phase for class
DaySummary.

kimtonyhyun:
Changed the function name. Try "methods(ds)" for a list of possible
functions of DaySummary,

-Tony

Am Mittwoch, 18. März 2015 schrieb forea :

> got an error.. what am I missing?
>
> >> ds = DaySummary('_data7/mouse7_day07_ego-left_ti2.txt', 'd7_ica001',
> 'excludeprobe')
>
> ds =
>
> DaySummary with properties:
>
> cells: [250x1 struct]
> trials: [100x1 struct]
> num_cells: 250
> num_trials: 100
> trial_indices: [100x4 int32]
>
> >> ds.plot_cell_phase(4);
> No appropriate method, property, or field plot_cell_phase for class
> DaySummary.
>
> On Wed, Mar 18, 2015 at 6:03 PM, Tony <notifications@github.com

kimtonyhyun:
@jmaxey Can you review this PR? I'd like to expand on it further -- but without adding too much to a single PR.

jmaxey:
Looking at it now.

jmaxey:
For the raster plot, do you normalize each trial? I do not see this in the code.

jmaxey:
If you are ok without having this normalization step, I can go ahead and merge the PR.

kimtonyhyun:
It looks ok without per-trial normalization so far,

-Tony (from phone)

jmaxey:
Ok. I know others in the lab normalize but but I guess when looking at one cell it doesn't make as much difference as when you are looking at multiple cells simultaneously. I'll merge it.

kimtonyhyun:
For different cells, we'll most certainly have to normalize.

PR #70, opened 2015-03-18
Kimth/reconst traces

kimtonyhyun:
@inanhkn With this PR, I propose to generate reconstructed outputs (filters / traces) into its own `rec_*.mat` file.

This approach, as opposed to computing reconstructions "on the fly" during `classify_ics` has the following advantages:

1. Reconstructions do not need to be repeatedly calculated.
2. We can elaborate on the reconstruction algorithm (e.g. allow for special handling if there are multiple blobs, as we have seen in a few cases). If the reconstruction were done on the fly during `classify_ics`, I would be less inclined for more complicated procedures.


inanhkn:
I agree.

PR #72, opened 2015-03-19
Raster for all cells during single trial

jmaxey:
@kimtonyhyun


kimtonyhyun:
Should be in `/view` directory. Also, can you show us a reference picture?

jmaxey:
example_plot

io/plot_trial_raster.m
```diff
@@ -0,0 +1,27 @@
+function [cellOrder] = plot_trial_raster(ic_traces,binned_frames,trial_num)
```
kimtonyhyun: `frame_indices` would be more general than `binned_frames`, since we may at some point run this function without time binning. (The requirement is that the frame indices match the length of traces.)

Also, add a brief comment about the function? The expected format of `ic_traces` (i.e. `[num_frames x num_ICs]`), etc. would be helpful.

jmaxey:
Ok cool. I wasn't sure where it should go. Do I need to move it locally and create a new PR?

jmaxey:
Sounds good. I'll make the changes.

jmaxey:
I'll also change it to just traces since the it could be plotting ic_traces or the fluorescence from the original movie isolated by the spatial ic_mask.

kimtonyhyun:
Good idea.

If you move the file on your local Git repository, and then push the changes up to the website ("remote") then it will accordingly update the PR. (You do not have to initiate a new one.)

io/plot_trial_raster.m
```diff
@@ -0,0 +1,27 @@
...
+cellOrder = idxN;
+
+for i=1:length(I)
+    new_norm_trial(i,:) = norm_trial(idxN(i),:);
+end
+
+new_norm_trial(:,binned_frames(trial_num,2)-binned_frames(trial_num,1)) = 1;
+new_norm_trial(:,binned_frames(trial_num,3)-binned_frames(trial_num,1)) = 1;
+
+figure(2)
```
forea: can you change this to just figure?

forea:
I also remember when you showed us these plots, you were able to use the cell order from one trial for another trial. Can we include that as a varargin?

jmaxey:
Great idea! Working on the change now.

PR #73, opened 2015-03-20
Motion correction using DFT

jmaxey:
@kimtonyhyun @forea @inanhkn 

Here is a script that will run the DFT motion correction. It is how I originally wrote it so only pulls a single frame at a time (vs the ~1000 frames that most of the functions now pull simultaneously).


reg/correctMotion_DFT.m
```diff
@@ -0,0 +1,64 @@
+function correctMotion_DFT(hdf5_dir,hdf5_in_name,hdf5_out_name)
+
+dataset = '/Data/Images';
+
+h5att = h5info(fullfile(hdf5_dir,hdf5_in_name),dataset);
```
kimtonyhyun: What's the reason for separating `hdf5_dir` and `hdf5_in_name`? For instance, in `hdf5_in_name`, you can use relative or absolute paths (e.g. `path/to/movie.hdf5`) that allows one to specify files in other directories.

reg/correctMotion_DFT.m
```diff
@@ -0,0 +1,64 @@
...
+imageStackSize = h5att.Dataspace.Size;
+rows = imageStackSize(1);
+cols = imageStackSize(2);
+frames = imageStackSize(3);
```
kimtonyhyun: FYI, in the `kimth/f2f-reg` branch, there is a helper function called [`get_dataset_info`](https://github.com/inanhkn/analysis/blob/kimth/f2f-reg/io/get_dataset_info.m) that performs the above fiew lines.

reg/correctMotion_DFT.m
```diff
@@ -0,0 +1,64 @@
...
+    iFFT = iFFT.*exp(i*2*pi*(-output(3)*Nr/nr-output(4)*Nc/nc));
+    regImageFFT = iFFT*exp(i*output(2));
+    regImage = abs(ifft2(regImageFFT));
+        
+    h5write(fullfile(hdf5_dir,hdf5_out_name),dataset,regImage,[1,1,f],[rows,cols,1]);
+end
+close(h);
+end
+
+function A_tr = mosaic_transform(A, ssm_filter, asm_filter)
```
kimtonyhyun: At this point, we should factor out `mosaic_transform`, since it is used by both the TurboReg and DFT registration methods. Not necessary for this PR, I can clean up afterwards.

reg/correctMotion_DFT.m
```diff
@@ -0,0 +1,64 @@
...
+im_ref = transform(baseImage);
+
+% Specify ROI
+%------------------------------------------------------------
+h_roi = figure;
+imagesc(im_ref); axis image; colormap gray;
+title('Select ROI');
+h_rect = imrect;
+mask_rect = round(getPosition(h_rect));
+cropBase = im_ref(mask_rect(2):mask_rect(2)+mask_rect(4),mask_rect(1):mask_rect(1)+mask_rect(3));
```
kimtonyhyun: Minor comment. Given that you use `mask_rect(2):...` here and also in the subsequent loop (Line 41), you should make the indices into its own variable.

reg/correctMotion_DFT.m
```diff
@@ -0,0 +1,64 @@
...
+    waitbar(f/frames);
+    image = h5read(fullfile(hdf5_dir,hdf5_in_name),dataset,[1,1,f],[rows,cols,1]);
+    cropImage = image(mask_rect(2):mask_rect(2)+mask_rect(4),mask_rect(1):mask_rect(1)+mask_rect(3));
+    tImage = transform(cropImage);
+    imageFFT = fft2(tImage);
+    [output, result] = dftregistration(baseFFT,imageFFT,100); 
+    
+    %%% Apply shift parameters to the non-blurred/cropped image
+    iFFT = fft2(image);
+    [nr,nc]=size(iFFT);
```
kimtonyhyun: Is `nr`, `nc` different from `rows`, `cols`?

reg/dftregistration.m
```diff
@@ -0,0 +1,212 @@
+
+
```
kimtonyhyun: Remove leading two newlines.

reg/correctMotion_DFT.m
```diff
@@ -0,0 +1,64 @@
+function correctMotion_DFT(hdf5_dir,hdf5_in_name,hdf5_out_name)
+
```
kimtonyhyun: Brief description of the function?

kimtonyhyun:
Left some minor comments. Once they are addressed, then will merge PR -- and I'll follow up with a refactor of `mosaic_transform`. I'll also then generate a few movies with the DFT registration so that we can get some sense of how it typically compares to TurboReg.

reg/correctMotion_DFT.m
jmaxey: I was doing this to match previous functions where the file name and directory were two separate inputs. I can change it.

kimtonyhyun: In your previous functions (like `concatenateHDF5`) I found myself padding with various `pwd`s for the directory input.

In general, I think it's nice to minimize the number of function parameters.

kimtonyhyun:
Note, there is a classification text file that should not be in the PR. Also, when the PR is ready review, then please give us a ping with the @ mention. (Otherwise, we don't know when PR has been updated.)

reg/dftregistration.m
jmaxey: @kimtonyhyun given that this file was downloaded directly from Mathworks website, I am inclined to not change anything in the file (even if it will clearly not affect the outcome).

jmaxey:
The classification file has been removed and it is ready to merge. @kimtonyhyun @forea @inanhkn

PR #74, opened 2015-03-25
Option to filter trials for raster plot in DaySummary

forea:
Similar to `plot_superposed_trials`, added option to filter for certain trials in `plot_cell_raster`, e.g. only "east" starts or only incorrect trials.

For example, from these three commands:
days{7}.plot_cell_raster(105)
days{7}.plot_cell_raster(105,'end','north');
days{7}.plot_cell_raster(105,'end','south');

I can get these three plots:
all_trials
north_trials
south_trials

### DIFF
--- a/io/DaySummary.m
+++ b/io/DaySummary.m
@@ -183,19 +183,42 @@ classdef DaySummary
             ylabel('Trace [a.u.]');
         end
         
-        function raster = plot_cell_raster(obj, cell_idx)
+        function raster = plot_cell_raster(obj, cell_idx, varargin)
+            % Optional arguments allow for filtering of trials, e.g.
+            %   "plot_cell_raster(cell_idx, 'start', 'east')"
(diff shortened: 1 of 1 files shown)


io/DaySummary.m
```diff
@@ -183,19 +183,42 @@ function plot_superposed_trials(obj, cell_idx, varargin)
...
+                    counter = counter+1;
+                    line = obj.get_trace(cell_idx, k);
+                    raster(counter,:) = interp1(linspace(0,1,length(line)),...
                                       line,...
                                       resample_grid,...
                                       'pchip');
-            end
+                end
+            end            
+            raster(all(~raster,2), : ) = []; %remove empty rows
```
kimtonyhyun: If you do, `raster = zeros(sum(display_trial), length(resample_grid));`, then you don't have to do the empty row removal.

PR #75, opened 2015-03-25
Factor out trial filtering

kimtonyhyun:
Factored out the trial filtering used in `plot_cell_raster` and `plot_superposed_trials`. Each of these methods now accept as its `varargin` a logical vector that indicates which trials to display.

There is a new method `filter_trials` that produces the logical vector. Specifically, what used to be
```
ds = DaySummary(...);
ds.plot_cell_raster(cell_idx, 'start', 'west');
```
is now performed as:
```
ds.plot_cell_raster(cell_idx, ds.filter_trials(`start`, `west`));
```

Note that `filter_trials` can apply multiple filters. For example `ds.filter_trials(`start`, `west`, `correct`)` will return the trials that start from the west arm _and_ the mouse correctly navigated to the goal arm.

### DIFF
--- a/io/DaySummary.m
+++ b/io/DaySummary.m
@@ -131,24 +131,35 @@ classdef DaySummary
             ylabel('Signal [a.u.]');
         end
         
-        function plot_superposed_trials(obj, cell_idx, varargin)
-            % Optional arguments allow for filtering of trials, e.g.
-            %   "plot_superposed_trials(cell_idx, 'start', 'east')"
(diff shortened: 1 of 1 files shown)


kimtonyhyun:
The only real advantage of the `filter_trials` refactor is that we don't have to duplicate code between `plot_cell_raster` and `plot_superposed_trials` (and other trial-based visualization functions we may introduce later).

PR #80, opened 2015-04-08
fix baseline in reconstructed traces

inanhkn:
Fixed the baseline offset and trend in the reconstructed traces. Refer to the below picture for the improvement.
untitled

@kimtonyhyun  I think you need to tell me once more what I need to modify to fix the NaNs in the filters. Is it the line where the filters are normalized? And if yes, how can the sum of an ICA filter be zero?

### DIFF
--- a/image_op/reconstruct_traces.m
+++ b/image_op/reconstruct_traces.m
@@ -145,10 +145,17 @@ end
 info.num_pairs = rec_filter_count; %#ok<STRNU>
 filters = filters(:,:,1:rec_filter_count);
 
+cells_to_exclude = [];
 for cell_idx = 1:rec_filter_count % Normalize
     filter = filters(:,:,cell_idx);
-    filters(:,:,cell_idx) = filter / sum(filter(:));
+    sum_filter = sum(filter(:));
+    if sum_filter>0
+        filters(:,:,cell_idx) = filter / sum_filter;
+    else
+        cells_to_exclude(end+1) = cell_idx;
+    end
 end
+filters(:,:,cells_to_exclude) = []; 
 
 % Reconstruct traces
 %------------------------------------------------------------
--- a/image_op/reconstruct_traces.m
+++ b/image_op/reconstruct_traces.m
@@ -1,4 +1,5 @@
-function reconstruct_traces(movie_source, ica_dir, varargin)
+%function reconstruct_traces(movie_source, ica_dir, varargin)
+
 % Reconstruct trace from the movie using the IC filters.
 %
 % inputs:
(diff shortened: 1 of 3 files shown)


inanhkn:
I still don't quite know how a thresholded ICA filter can be all zeros,but I guess I fixed the NaN problem.

inanhkn:
ICs that are hot pixels should now be eliminated. I also set the mad threshold for event detection to a higher value. The latter should be addressed more comprehensively in the future, preferrably providing a way for user to set the threshold on the fly( Possibly to be implemented by @kimtonyhyun ).

image_op/reconstruct_traces.m
```diff
@@ -144,13 +148,21 @@ function reconstruct_traces(movie_source, ica_dir, varargin)
...
     filter = filters(:,:,cell_idx);
-    filters(:,:,cell_idx) = filter / sum(filter(:));
+    sum_filter = sum(filter(:));
+    if sum_filter>0 && sum(filter(:)>0)>=min_num_pixels
+        filters(:,:,cell_idx) = filter / sum_filter;
+    else
+        cells_to_exclude(end+1) = cell_idx; 
+    end
 end
+filters(:,:,cells_to_exclude) = [];
```
kimtonyhyun: If we're removing some filters this way, we then need to readjust `rec_filter_count` as:
```
rec_filter_count = rec_filter_count - length(cells_to_exclude);
info.num_pairs = rec_filter_count;
```

image_op/reconstruct_traces.m
```diff
@@ -171,7 +183,11 @@ function reconstruct_traces(movie_source, ica_dir, varargin)
     pix_active = find(rec_filter>0);
     movie_portion = M(pix_active,:)';
     movie_portion(movie_portion<threshMov) = 0;
-    traces(:,cell_idx) = movie_portion * rec_filter(pix_active);  
+    trace_this = (movie_portion * rec_filter(pix_active))';  
+    % Fix baseline
```
kimtonyhyun: These operations for baseline removal are not obvious, can you explain in comments what it is doing here?

image_op/reconstruct_traces.m
```diff
@@ -31,8 +32,11 @@ function reconstruct_traces(movie_source, ica_dir, varargin)
 % Hakan Inan (Mar 15)
 %
 
+% Minimum number of active pixels in an IC filter to be regarded as valid
+min_num_pixels = 4;
+
 % Reconstruction parameters
-threshMov = -1; % By default keep everything
+threshMov = 0;
```
kimtonyhyun: Can we leave `threshMov = -1`. Setting `threshMov = 0` arbitrarily removes negative values from the movie, which is not the most conservative action. You can continue to use negative removal for your own reconstructed traces.

image_op/reconstruct_traces.m
kimtonyhyun: In fact, it would be better to factor out the baseline removal to a function of its own, in the `signal_op` directory.

inanhkn:
Made the changes.

PR #83, opened 2015-04-15
better method for baseline removal

inanhkn:
After observing remaining baseline fluctuations in many traces, I wrote a better routine for the baseline removal. It is simple, but requires a bit more computation. Refer to the function for the details.
Here are some examples to visualize the improvement: (the blue traces are the ones obtained after baseline removal with the old procedure, orange traces are the fitted baselines with the new routine.)

### DIFF
--- a/signal_op/fix_baseline.m
+++ b/signal_op/fix_baseline.m
@@ -1,4 +1,4 @@
-function trace_out = fix_baseline(trace_in)
+% function trace_out = fix_baseline(trace_in)
 % Remove baseline offset of cell trace with following steps:
 %    - sample baseline for overlapping small segments
 %    - smooth baseline samples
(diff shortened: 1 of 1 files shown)


kimtonyhyun:
One comment, `trace_in` currently needs to be `[num_samples x 1]` vector. It would be nice if it also worked for `[1 x num_samples]`.

kimtonyhyun:
Also, `trace_out` has trailing `NaN`s.

inanhkn:
Fixed the issues. You can input a row vector, but it'll come out as a column vector.

kimtonyhyun:
Looks good.

signal_op/fix_baseline.m
```diff
@@ -1,8 +1,48 @@
...
+        filt_out(idx_sample+1) = mean(b);
+    end
+
+    % Interpolate the sample baseline points
+    subt = interp1(baselines(:,1),filt_out,(1:num_frames)','spline');
+
+    trace_out = trace_in-subt;
+
+    
+    plot(trace_in)
```
kimtonyhyun: Actually, you should remove these 'plot' commands.

PR #86, opened 2015-04-17
Medfilt for removing hot pixels

kimtonyhyun:
The function `medfilt_movie` performs 2D median filtering on a frame-by-frame basis in order to eliminate hot pixels. The syntax is the same as the other file-to-file preprocessing functions, i.e. `medfilt_movie(movie_in, movie_out)`.

The motivation comes from Cohort 9 Mouse 2. For whatever reason, this particular microscope has a tremendous number of hot pixels in the camera. (An aside @forea, we should get this fixed by Inscopix.)

The hot pixels are stationary with respect to the camera which means that, following motion correction, these pixels move around, giving rise to artifactual ICs. For c9m2, my current idea is to apply median filtering right after concatenation.

The effect of median filtering can be seen here: [medfilt_demo.mat (400 MB)](https://www.dropbox.com/s/n1guzcwx4tw5p13/medfilt_demo.mat?dl=0)

When you load `medfilt_demo.mat` into the Matlab workspace, you'll find two movie snippets from c9m2: `M` and `M_medfilt`. The first movie is the raw concatenated input to `medfilt_movie` and the latter is the output. You can view these snippets as usual via `view_movie(M, 'repeat', 5)`. It should be evident that some very obvious hot pixels are eliminated by the median filtering.

### DIFF
--- a/image_op/medfilt_movie.m
+++ b/image_op/medfilt_movie.m
@@ -1,4 +1,4 @@
-function medfilt_movie(movie_in, movie_out, varargin)
+function medfilt_movie(movie_in, varargin)
 % Apply median filtering from file ('movie_in') to file ('movie_out').
 % Optional argument specifies the half-width of the median filter.
 %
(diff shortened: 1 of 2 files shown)


inanhkn:
Just an idea, wouldn't a hot pixel have a very high mean value along the time axis? If so, then we can identify all the hot pixels in the beginning and replace them with their neighborhood average at each frame, avoiding filtering the whole thing.

I may as well be missing something.

inanhkn:
This is given that the camera has the hot pixels at the same pixel location, independent of the motion.

kimtonyhyun:
Yeah, the hot pixels needs to be evaluated before motion correction, since they are fixed in place with respect to the camera. (And they'll start moving once MC has been applied.)

Totally with you on the initial identification of the trouble regions, so that we don't apply filtering on the entire image. If you recall, this is the way I treated some hot pixels in the past for c9m7. On the other hand, c9m2 has so many hot pixels in the image -- it's crazy. Take a look at the videos that I sent (`medfilt_demo.mat`).

kimtonyhyun:
Another approach to identifying hot (or "stuck") pixels would be to look at their variance over time -- my guess is that they may show reduced variance compared to normal pixels.

kimtonyhyun:
Closing in favor of #90

kimtonyhyun:
@inanhkn For your use

inanhkn:
merging it.

PR #90, opened 2015-05-11
remove hot pixels

inanhkn:
The big computer won't let you go Tony..

With this script the hot pixels should not be a problem anymore. The script identifies the hot pixels and removes them by median filtering only those pixels. 
Example call: 
`remove_hot_pixels('c9m7d12.hdf5','');`

Important note: This script should be called before the motion correction is done since the hot pixels will start moving around after the motion correction.

### DIFF
--- a/image_op/remove_hot_pixels.m
+++ b/image_op/remove_hot_pixels.m
@@ -1,0 +1,79 @@
+function remove_hot_pixels(movie_in,movie_out,varargin)
+% Remove hot pixels in the raw movie.
+%
+% Input arguments:
+%   movie_in:  Name of incoming HDF5 movie
+%   movie_out: Name of outgoing HDF5 movie 
+%
+% Optional input argument:
+%   num_hot_pixels: Maximum estimated number of hot pixels. Default is 30.
+%
+% Example usage:
+%   remove_hot_pixels('c9m7d12.hdf5','');
+%
+
+if ~isempty(varargin)
+    num_hot_pixels = varargin{1};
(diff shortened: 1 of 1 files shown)


kimtonyhyun:
This works well enough for a few hot pixels (I tried on `c9m6d12`). On the other hand, it doesn't solve the issues for the Miniscope that has a ton of hot pixels (i.e. Mouse 2). Maybe we need to treat `c9m2` as a special case... I'll show you next time you are at Herrin.

kimtonyhyun:
I bet the big computer has some sort of a global user setting, since I installed Github for Windows... Let me see if I can undo that.

inanhkn:
Tony, did you set the maximum number of hot pixels to a higher value? You can set it to 1000 and it will remove exactly 1000 'hottest' pixels (the default is 30:) ). It will still be infinitely faster than median filtering all [height]x[width] number of pixels.

inanhkn:
Also, if you think there are movies with many hot pixels adjacent to one another, let me know. I naively assumed that would not be the case. I can make changes that take that into account.

kimtonyhyun:
Actually, with increased number of hot pixels (1500 for `c9m2d05`), the output movie looks much better. The PR looks good to me.

kimtonyhyun:
Btw, I'm processing c9m2 now on the old big computer (`LY-imaging.stanford.edu`) using this filtering as the first step in the pipeline (after concatenation).

inanhkn:
I have made a few minor additions that does not change the functionality. I'm merging it.

PR #91, opened 2015-05-20
added LS

inanhkn:
This PR deals with the addition of the Least Squares reconstruction of the traces from the movie and the filters. The script now reconstruct traces by using least squares by default. Call it with the extra parameter `use_legacy` to perform applying mask to the movie for extraction.

More explicitly, this example call results in LS extraction:
`reconstruct_traces('c9m7d08_cr_mc_cr_norm_dff_ti2.hdf5','')`

This one results in extraction by applying mask:
`reconstruct_traces('c9m7d08_cr_mc_cr_norm_dff_ti2.hdf5','','use_legacy')`

### DIFF
--- a/image_op/reconstruct_traces.m
+++ b/image_op/reconstruct_traces.m
@@ -201,11 +201,14 @@ else
     F = reshape(filters,height*width,rec_filter_count);
     idx_nonzero = find(sum(F,2)>0);
     F_small = F(idx_nonzero,:); 
-    proj_F = (F_small'*F_small)\F_small';
-
-    for idx_frame = 1:num_frames
-        traces(idx_frame,:) = fix_baseline( proj_F *M(idx_nonzero,idx_frame) );
+    
+    % Caution: Below line requires memory of size roughly 10-20% of M
+    traces = (F_small'*F_small)\(F_small'*M(idx_nonzero,:)); 
+    traces = traces';
+    for k = 1:rec_filter_count;
+        traces(:,k) = fix_baseline(traces(:,k));
     end
+
 end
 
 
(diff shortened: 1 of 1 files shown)


image_op/reconstruct_traces.m
```diff
@@ -182,15 +185,33 @@ function reconstruct_traces(movie_source, ica_dir, varargin)
...
+    %Least Squares Method
+    F = reshape(filters,height*width,rec_filter_count);
+    idx_nonzero = find(sum(F,2)>0);
+    F_small = F(idx_nonzero,:); 
+    
+    % Caution: Below line requires memory of size roughly 10-20% of M
+    traces = (F_small'*F_small)\(F_small'*M(idx_nonzero,:)); 
+    traces = traces';
+    for k = 1:rec_filter_count;
+        traces(:,k) = fix_baseline(traces(:,k));
```
kimtonyhyun: You can pull `fix_baseline` out of the `use_legacy` if-then statement, since it's common to both branches

kimtonyhyun:
I checked out the traces coming from the revised method on c9m7d18 -- looks good to me.

kimtonyhyun:
I've been examining the performance of LS for de-crosstalking overlapping cells. It's an improvement from before (simple projection), but not a dramatic one.

For example, below are two overlapping cells where the top panel shows traces extracted by LS, and the bottom panel shows traces extracted by simple projection:
20150522_ls-perf

LS does show better de-crosstalk performance (ratio of red to green amplitudes), but qualitatively the trace pairs are similar in the LS / ROI projection cases.

kimtonyhyun:
Here's also a case where the LS performs much better at decrosstalk (same pair as above):
20150522_ls-perf2

inanhkn:
I fixed the redundancy you mentioned, I'm merging it.

PR #92, opened 2015-05-24
fix minor bug in reconstruct_traces

inanhkn:
It turns out I forgot to save the latest script before merging `#91`.  I've committed the latest change.


inanhkn:
I'm merging this as well assuming I didn't skip another stupid mistake:)

PR #93, opened 2015-05-26
Kimth/bin traces

kimtonyhyun:
The function `bin_traces` will temporally bin traces in a way that's consistent with our existing time binning methods (e.g. `bin_movie_in_time`). Example usage:

```
>> orig_indices = get_trial_frame_indices('_data/c9m7d06.txt');
>> binned_traces = bin_traces(traces, orig_indices, 2);
```

Note that `bin_traces` takes the original, _unbinned_ frame indices as an argument!

This was necessary because the EM analysis for c9m7d06 was apparently performed for the full frame (20 Hz) movie, whereas we wanted to compare the EM result to our standard 10 Hz analyses.

I also added an `assert` during initialization of `DaySummary` to make sure that the length of traces is consistent with the number of frames as indicated by the maze text file.

### DIFF
--- a/io/DaySummary.m
+++ b/io/DaySummary.m
@@ -55,7 +55,12 @@ classdef DaySummary
             %   TODO: Bring in centroids corresponding to mouse position
             %------------------------------------------------------------
             [trial_indices, loc_info, trial_durations] =...
-                parse_plusmaze(plusmaze_txt);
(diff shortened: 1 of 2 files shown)


kimtonyhyun:
@inanhkn Can you review?

PR #94, opened 2015-05-28
Baseline toggle

kimtonyhyun:
This PR adds the ability to perform baseline correction on traces during manual cell checking (using the existing `signal_op/fix_baseline`). It can be invoked by typing `b` during the movie viewer:
```
Classifier (211/541) >> c
Cell viewer >> b
  Showing trace with baseline correction
Cell viewer >> 
```
(It can also be toggled off, by typing `b` once more.)

The motivation was that CellMax traces have not been explicitly corrected for baseline wander.

Here are example traces before and after.

Before baseline removal:
before

After baseline removal:
after

### DIFF
--- a/classification/support/view_cell_interactively.m
+++ b/classification/support/view_cell_interactively.m
@@ -7,8 +7,9 @@ function [resp, movie_clim] = view_cell_interactively(ds, cell_idx, movie, fps, 
 %
 
 filter = ds.cells(cell_idx).im;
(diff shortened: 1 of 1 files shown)


forea:
tested it out, works for me

kimtonyhyun:
Btw, this toggle really shows that `fix_baseline` is doing a very nice job.

PR #96, opened 2015-06-18
Replace min projection with mean projection in remove_hot_pixels

inanhkn:
Replacing min projection with mean projection in remove_hot_pixels seems to yield much better results in hot pixel removal. With this PR I propose this change.


kimtonyhyun:
Can you give more details on the expected improvement of this change (e.g. some example movie snippets)?

PR #101, opened 2015-08-13
Inanhkn/extract cells

inanhkn:
The cell extraction routine that first finds filters solving the source separation problem and then finds traces with Least Squares.  Call it as:

`extract_cells('movie file path','pca*.mat file path',[maximum estimated number of cells])`

The script outputs a `rec*.mat` file that contains filters and traces.


inanhkn:
The script now outputs something called inverse quality and plots it. It looks like this:
inverse_qualities

It might help the user when to stop examining cells in manual cell checking process. The bigger the value is, the worse the cell is considered to be by the cell extractor.

pca_ica/extract_cells.m
```diff
@@ -0,0 +1,326 @@
...
+fprintf('%s : Loading movie singular vectors...\n',datestr(now));
+
+load(pca_source,'pca_info','pca_filters');
+is_trimmed = pca_info.trim.enabled;
+h = pca_info.movie_height;
+w = pca_info.movie_width;
+t = pca_info.movie_frames;
+
+if is_trimmed
+    load(pca_source,'idx_kept');
```
kimtonyhyun: This needs to be updated, to be consistent with PCA storage scheme of #104.

pca_ica/extract_cells.m
```diff
@@ -0,0 +1,326 @@
...
+fprintf('%s : Extracting traces...\n',datestr(now));
+M = reshape(M,h*w,t);
+idx_nonzero = find(sum(F,2)>0);
+M_small = M(idx_nonzero,:);
+F_small = F(idx_nonzero,:);
+T = (F_small'*F_small+0*eye(size(F,2))) \  (F_small'*M_small);
+clear M;
+
+% Reconstruction settings
+info.type = 'simple';
```
kimtonyhyun: We need to come up with a different name here.

pca_ica/extract_cells.m
```diff
@@ -0,0 +1,326 @@
...
+%       found potential is as far as the extraction algorithm is concerned.
+%       It is a [1 x # of extracted cells] array whose entries are
+%       normalized to [(1/# of pixels),1] range. This array is also plotted
+%       on the screen at the end of execution.
+%
+%   Hakan Inan (Aug 2015)
+%
+
+% Some parameters
+maxsize_chunk_GPU = 500;
```
kimtonyhyun: Brief comments about what a "chunk" refers to would be helpful.

pca_ica/extract_cells.m
```diff
@@ -0,0 +1,326 @@
...
+maxsize_chunk_CPU = 3000;
+corr_thresh = 0.1;
+filter_thresh = 0.25;
+
+max_num = min(max_num,1000); % Internal control for max_num
```
kimtonyhyun: Generally, if you are overwriting user-specified parameters, then it is common courtesy to indicate the override (e.g. by console output).

pca_ica/extract_cells.m
```diff
@@ -0,0 +1,326 @@
...
+
+fprintf('%s : Loading movie singular vectors...\n',datestr(now));
```
kimtonyhyun: Can we use '%s: ...' (no space after timestamp) for consistency the rest of the codebase?

kimtonyhyun:
Fyi, runtime on my PC:
```
>> extract_cells(sources.miniscope, sources.pca, 600);
20-Aug-2015 21:22:57 : Loading movie singular vectors...
20-Aug-2015 21:23:02 : Extracting filters...
	 	 	 Extracted 600 potential cells 
20-Aug-2015 21:56:08 : Cleaning filters and removing duplicates...
	 	 	 1 cells were merged into others 
20-Aug-2015 21:56:27 : Loading movie for trace extraction...
20-Aug-2015 21:57:26 : Extracting traces...
20-Aug-2015 21:57:57: Writing the output to file... 
20-Aug-2015 21:57:57: Output will be saved in rec_150820-215757.mat 
20-Aug-2015 21:58:25 : All done! 
```

inanhkn:
I'm going to work on `max_chunksize_GPU` and `max_chunksize_CPU` as they are determined very arbitrarily right now; I'll also add proper commentary. As for the `info.type` I propose to have the names of the steps involved, namely PCA and sparse source separation(SSS for short). So, it could be `'PCA+SSS'`. 

I am very surprised with the runtime for the matrix multiplication (in the filter extraction step). One bad thing about it is that the runtime is quadratic in the number of pixels (trimming reduces runtime by a factor of 4). One does not need to scan through all the pixels, but I want to keep it as is to be safe. I'll make `max_chunksize_CPU` adaptive and make more extensive use of the computer memory to make it faster for slower machines. Is your PC the lenovo laptop?

kimtonyhyun:
`PCA+SSS` seems like a good name.

I generated a new PCA file for c11m1d12, using no trimming, but with the medfilt option on (700 PCs).

Actually, "my PC" is the the desktop that I moved back home. It's a 2010 Dell (fairly high end at the time; e.g. Core i7). I'd expect it to be pretty comparable to the desktop that I use at Herrin (the refurbished one from MarkW).

inanhkn:
I see. It takes the old 'big' computer at at Herrin ~15min to do that step with ~250k pixels (2x binned frame)  without trimming. It could be a processor speed issue, both my computer and the old big computer have 3.5Ghz processors. It takes my computer ~6min to do the same processing though..

kimtonyhyun:
Hooray for Intel ;)

kimtonyhyun:
What happened to the diff? Why is the file being displayed as binary, rather than text?

inanhkn:
I really don't know. Is there a way to convert the display back to text?

kimtonyhyun:
Never happened to me before. In the worst case, you can create a new branch, copy the file over, and open a new PR.

kimtonyhyun:
Let's try copying the file to a new PR, to get the text diff back. That way, it becomes possible to code review.

I would also suggest not applying `fix_baseline` inside the extraction function. When classifying cells (where we have the option of applying -- or not -- the baseline correction), I've noticed that the baseline removal can have non-ideal effects on certain types of traces (e.g. very active ones).

inanhkn:
I have some other fundamental changes in mind; I'll make them also considering your suggestions from here. I'll continue in a new branch and make another pull request when ready. In the meanwhile I'm closing this one.

PR #104, opened 2015-08-17
add trimming of pixels

inanhkn:
This PR introduces a new proposed feature for PCA computation, namely trimming of pixels in the movie that are not active.  You can call `run_pca` now as

`run_pca('movie source', #of PCs,'trim');`

To have the script trim half of the movie pixels (ones whose maximums are lower than the median of the max projection image) prior to computing PCA. Assuming that cells occupy less than half of the frame pixels this is an OK way to get rid of pixels. (I have even observed increase in the quality of cell extraction output by eliminating pixels, possibly due to removing the already not-so-relevant variance)

The pixels indices that are kept are stored in the output `.mat` file as the variable `idx_kept`. Whether trimming is done to a `pca*.mat` file can be seen by checking a new property `is_trimmed` under the variable `pca_info`.

This feature speeds up covariance matrix calculation prior to PCA and also ultimately reduces the computation in a subsequent cell extraction step.

### DIFF
--- a/pca_ica/run_pca.m
+++ b/pca_ica/run_pca.m
@@ -12,7 +12,7 @@ function run_pca(movie_source, num_PCs,varargin)
 %       values.
 %
 % Example usage:
-%   run_pca('c9m7d25_dff.hdf5', 500,1);
+%   run_pca('c9m7d25_dff.hdf5', 500,'trim');
 %
 
 % Defaults
@@ -58,10 +58,11 @@ pca_info.movie_width  = width;
 pca_info.movie_frames = num_frames;
 pca_info.num_PCs = num_PCs; 
 pca_info.trimmed = do_trim; %#ok<STRNU>
-save(savename, 'pca_info', 'pca_filters', 'pca_traces', 'S');
 
 if do_trim
-    save(savename,'idx_kept');
+    save(savename,'pca_info', 'pca_filters', 'pca_traces', 'S','idx_kept');
+else
+    save(savename, 'pca_info', 'pca_filters', 'pca_traces', 'S');
 end
 
 fprintf('%s: All done!\n', datestr(now));
--- a/pca_ica/run_pca.m
+++ b/pca_ica/run_pca.m
@@ -57,7 +57,7 @@ pca_info.movie_height = height;
 pca_info.movie_width  = width;
 pca_info.movie_frames = num_frames;
 pca_info.num_PCs = num_PCs; 
-pca_info.trimmed = do_trim; %#ok<STRNU>
+pca_info.is_trimmed = do_trim; %#ok<STRNU>
 
 if do_trim
     save(savename,'pca_info', 'pca_filters', 'pca_traces', 'S','idx_kept');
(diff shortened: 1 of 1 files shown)


inanhkn:
With the latest modification, now you can run median filtering with `run_pca.m` without having to execute `medfilt_movie` it as a separate step and have another movie matrix. 

Call `run_pca` as

`run_pca('movie source', #of PCs,'trim','medfilt');`
to first median filter the movie matrix and then trim it prior to PCA. Addition of median filtering has no extra memory load since it does filtering on per-frame basis. It took my computer 2 mins to median filter a [484x568x19414] movie matrix.

pca_ica/run_pca.m
```diff
@@ -1,4 +1,4 @@
-function run_pca(movie_source, num_PCs)
+function run_pca(movie_source, num_PCs,varargin)
```
kimtonyhyun: As a style comment, let's have an extra space after each comma

pca_ica/run_pca.m
```diff
@@ -33,7 +81,19 @@ function run_pca(movie_source, num_PCs)
...
-pca_info.num_PCs = num_PCs; %#ok<STRNU>
-save(savename, 'pca_info', 'pca_filters', 'pca_traces', 'S');
+pca_info.num_PCs = num_PCs; 
+
+pca_info.is_trimmed = do_trim; 
+
+pca_info.is_medianfiltered = do_medfilt;  %#ok<*STRNU>
+if do_medfilt
+    pca_info.medfilt_halfwidth = medfilt_halfwidth;
+end
```
kimtonyhyun: Until we come up with a better ideal, I'd suggest a hierarchical storage scheme for `pca_info`, like:
```
pca_info.movie_height = height;
...
pca_info.num_PCs = num_PCs;

pca_info.trim.enabled = ...
pca_info.trim.idx_kept = ...

pca_info.medfilt.enabled = ...
pca_info.medfilt.halfwidth = ...
```

and also to use `idx_kept = ones(num_pixels,1)` (use column or row vector, to be consistent with the `idx_kept` returned by the `find` operation).

This way, we don't have to worry about different variables (e.g. `idx_kept`, in the current implementation) being present in the PCA `.mat` file.

pca_ica/run_pca.m
```diff
@@ -6,24 +6,72 @@ function run_pca(movie_source, num_PCs)
 %   movie_source: Movie filename
 %   num_PCs: Number of requested PCs
 %
+% Variable input arguments:
+%   'trim': if added as an argument then the script trims pixels in the 
+%       movie that do not cross above the median of the maximum pixel 
+%       values.
+%   'medfilt': Add it as an argument to perform median-filtering on the 
+%       movie on a per-frame basis before PCA. Highly recommended since
+%       especially the miniscope camera might have various artifacts.
```
kimtonyhyun: I think you can leave out the "Highly recommended since..." in this comment block.

pca_ica/run_pca.m
```diff
@@ -6,24 +6,72 @@ function run_pca(movie_source, num_PCs)
...
 % Example usage:
-%   run_pca('c9m7d25_dff.hdf5', 500);
+%   run_pca('c9m7d25_dff.hdf5', 500,'trim','medfilt');
 %
 
+% Defaults
+do_trim = 0;
+medfilt = 0;
+
+if ~isempty(varargin)
```
kimtonyhyun: You could just do:
```
for k = 1:length(varargin)
    switch varargin{k}
        ...
    end
end
```

inanhkn:
I made the suggested modifications.

PR #106, opened 2015-08-20
Kimth/deprecate ica

kimtonyhyun:
None of us are using ICA filters/traces anymore. I'd like to propose making `reconst` the default from now on. This is the beginning of some code renovation for Cohort 11 analysis.

For example, instead of writing:
```
M = classify_cells(sources, 'rec001', 'reconst');
m1d12 = DaySummary(sources.maze, 'rec001', 'reconst');
```
you'd now just write:
```
M = classify_cells(sources, 'rec001');
m1d12 = DaySummary(sources.maze, 'rec001');
```


inanhkn:
I agree to the changes.

PR #107, opened 2015-08-23
Kimth/docs

kimtonyhyun:
Going forward, I propose that we explicitly describe the expected contents of each mat file, e.g. PCA, reconstruction files, in the `docs` directory.

Here is the initial file to get us started. I'll follow up by cleaning up the actual contents of each `rec_*.mat` file. (It's currently a mess.)


kimtonyhyun:
Hit `View` to see the document with MD formatting:
view

PR #109, opened 2015-08-25
Inanhkn/pca sss

inanhkn:
This PR is meant to be the continuation of #101 which introduced the cell extraction method with PCA+SSS. I made some changes since then:

1) `run_pca` is updated. The script now estimates an upper bound on the number of cells in a movie using a newly added script `local_maxima_2D.m`. If user still wants to manually specify number of PCs, s/he can do so by calling the script as
`run_pca('movie source', 'num_PCs',<number of asked PCs>)` .
To let the script decide, just type
`run_pca('movie source')` .

I would highly recommend median filtering and trimming with `run_pca`. In that case, type
`run_pca(movie_source, 'medfilt','trim')` .

2)  `local_maxima_2D.m` is added. This script takes an image (in our case, the max-projection image) and detects the local maxima in it. It estimates an upper bound on the number of local maxima (cells in our case) by computing 2 goodness measures: laplacian and pixel intensity at the found centroids(local maxima). The (scaled) laplacian and intensity values are then sorted and an exponential is fit to the these trends, assuming(confidently) that these trends obey power law. The time constants of the exponentials are extracted and the upper bound is calculated as the average of 4 times the time constants.
You can call   `local_maxima_2D.m` as 
`centroids = local_maxima_2D(<the image>,'visualize')`
to get the centroids and plot the image with  the centroid overlay + the trends for laplacian and pixel intensities with the exponential fit. Below are example output plots:
img
trends

After running PCA, call extract cells by typing
`extract_cells('movie file path','pca*.mat file path')` .
Refer to #101 for more details on `extract_cells` script.


inanhkn:
One modification is that per @kimtonyhyun 's suggestion, the baseline removal from the traces are now optional. To remove the baseline from the traces, now call `extract_cells` as
`extract_cells(<movie file path>,<pca*.mat file path>,'fix_baseline') ;` .

inanhkn:
By the way, `extract_cells` is still displayed as a binary file, I'll look for a way to change the display to text.

pca_ica/run_pca.m
```diff
@@ -54,6 +63,22 @@ function run_pca(movie_source, num_PCs, varargin)
...
+end
+
+% Detect local maxima in max-projection image to get an estimated maximum 
+%number of cells in the movie
+if autoset_num_PCs
+    cents = local_maxima_2D(max_proj);
+    num_PCs = size(cents,2);
+    if num_PCs>max_num_PCs
+        num_PCs = max_num_PCs;
+    end
```
kimtonyhyun: Can we add a similar safeguard for `min_num_PCs`, e.g. 400?

pca_ica/run_pca.m
```diff
@@ -54,6 +63,22 @@ function run_pca(movie_source, num_PCs, varargin)
...
+if do_trim || autoset_num_PCs
+    max_proj = max(M,[],3);
+end
+
+% Detect local maxima in max-projection image to get an estimated maximum 
+%number of cells in the movie
+if autoset_num_PCs
+    cents = local_maxima_2D(max_proj);
+    num_PCs = size(cents,2);
+    if num_PCs>max_num_PCs
```
kimtonyhyun: In such cases, it would be helpful to write to console that the automatically determined number of PCs has been overwritten, since it may indicate that the movie is somewhat weird.

kimtonyhyun:
Also the "Example usage" comments of `run_pca.m` needs to be updated.

kimtonyhyun:
What's with the `.gitattributes`?

kimtonyhyun:
While we're changing things in `run_pca`, can we also rename `pca_info`, `pca_filters`, `pca_traces` into just `info`, `filters`, `traces`? This would mean that the variable structure within a PCA mat file is identical to that of a "Rec" mat file.

kimtonyhyun:
Can you comment on the "inverse quality metric" provided by `extract_cells.m`? Does it, for example, consider the trace, the filter, or both?

kimtonyhyun:
Also, if you're going to go through the trouble of computing a "quality metric", then it would be nicer to plot it non-inverted.

kimtonyhyun:
Some initial observations on c11m1d12 (which I've looked at multiple times now with various cell extraction methods, e.g. CellMax under different conditions).

I ran PCA as follows `run_pca(..., 'medfilt')`, i.e. no trim, and allowed for auto determination of `num_PCs` (came to 745).

SSS (`extract_cells`) returned 911 candidates. I classified up to 500, since the yield was becoming very low by this point. Of the ones I manually looked at, I classified 264 to be cells. Here is the current cell map:
c11m1d12_sss

My high level observations are that:
- There are many very small filters -- peppered especially along the edges of the FOV -- which can likely be eliminated based on a size requirement.
- In a minority -- but notable -- number of cases, the filter boundary was offset from the actual cell blobs in the movie. I was generally tolerant of these filters, but in a few cases the offset was significant enough for me to reject the cell, though I could see that the filter had captured part of the cell in the movie (and was not splitting it with another filter).
- In general, I did not see too many overlapping cells, even in the case of offset filters (mentioned above) -- so it did not seem to me that the filters were displaced because it was competing with another nearby filter for a single cell in the movie.

Note that there are a number of decently sized filters in c11m1d12_sss (i.e. not the really tiny speckle) that have a red boundary in the cellmap above. It could be the case that I didn't get to them within the first 500, or that I rejected them based on the quality of the traces -- to be determined.

I also compared the results against CellMax on c11m1d12, which yielded 424 classified cells out of 888 candidates. Blue boundaries are CellMax, red boundaries are SSS:
c11m1d12_cm-vs-sss

My impression from this comparison is that the two methods have a very high overlap, but that CellMax has a significantly higher yield (with the caveat that I can't -- at the moment -- replicate the these results with the latest variant of the CM code). I will have to validate whether some of the cells that were rejected in SSS (especially near the center of the FOV) was simply because I didn't get to them in my classification. Further dissection to follow.

kimtonyhyun:
Here is the cell map where I went through all 911 candidates (there were only 1 or 2 cells in the range 500+):
c11m1d12_sss2

inanhkn:
inverse quality metric is based only on the filters. It is inverted by its nature; I don't think de-inverting it would gain the user more insight. I'd like to keep it as it is.

inanhkn:
In the near future I'm considering to add another metric based on traces, I'll probably drop the plotting of this current metric then.

kimtonyhyun:
Note that this PR breaks `run_ica.m` since the internal structure of the PCA mat file is changed.

PR #111, opened 2015-08-27
Kimth/classify commands

kimtonyhyun:
Addition of a few simple commands in `classify_cells` to speed up the process:
- Can now type `p!` or `c!` to classify a cell to be phase-sensitive or cell, without viewing the trace with the movie.
- Can now press [enter] (i.e. empty string) in order to increment the candidate index.


inanhkn:
Great.

PR #115, opened 2015-08-28
add option to change medfilt halfwidth

inanhkn:
Added option to set median filter halfwidth when calling `run_pca`. You can now call it as:
`run_pca(<'movie source'>,'medfilt',X);`
to set the median filter halfwidth to X.

### DIFF
--- a/pca_ica/run_pca.m
+++ b/pca_ica/run_pca.m
@@ -11,7 +11,9 @@ function run_pca(movie_source, varargin)
 %       movie that do not cross above the median of the maximum pixel 
 %       values.
 %   'medfilt': Add it as an argument to perform median-filtering on the 
-%       movie on a per-frame basis before PCA. 
+%       movie on a per-frame basis before PCA. Use input pair ('medfilt',X)
+%       to set the median filter halfwidth to X. X must be a positive 
+%       integer.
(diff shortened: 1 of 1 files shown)


pca_ica/run_pca.m
```diff
@@ -34,6 +36,12 @@ function run_pca(movie_source, varargin)
                 do_trim = 1;
             case 'medfilt'
                 do_medfilt = 1;
+                if k<length(varargin)
```
kimtonyhyun: If `medfilt_halfwidth` is a mandatory input that follows the `'medfilt'` option, then you should write:
```
medfilt_halfwidth = varargin{k+1};
```
(i.e. simply assume that the variable is coming in).

inanhkn:
It's not mandatory. User can still do `'medfilt'` and the script will use halfwidth of 1 (default).

pca_ica/run_pca.m
```diff
@@ -25,7 +27,7 @@ function run_pca(movie_source, varargin)
 medfilt_halfwidth = 1;
 autoset_num_PCs = 1;
 max_num_PCs = 1000;
-min_num_PCs = 300;
+min_num_PCs = 300;
```
kimtonyhyun: I'm going to merge this for now, but stylistically, let's make sure there are no trailing whitespace in merged code.

PR #116, opened 2015-08-28
Kimth/classify map

kimtonyhyun:
This has two features, pertaining to `classify_cells.m`:

__1. Slow entry into `view_cell_interactively.m`:__

In more recent cell extraction runs, where there are ~1000 cell candidates, I've noticed a significant delay when entering the "view_cell_interactively" portion (i.e. watch traces along with the movie) of the classification procedure.

This has to do with a large number of calls to the Matlab function `text`, which was being used to label cell indices associated with each outline over the movie. (It was hidden by default, shows up when "n" was pressed. I've used it fairly sparingly in my classification sessions.) For now, I just disabled the index labeling, and the entry speed into "view_cell_interactively" has recovered.

__2. Ability to view the cell map during the classification procedure:__

During cell classification, you can now type "m" (for "map"), as follows:
```
>> sources = data_sources

sources = 

          pca: '_data/pca_n700.mat'
          fps: 10

>> M = classify_cells(sources, 'sss1_new');
  27-Aug-2015 18:33:00: Loading "_data/c11m1d12_cr_gfix_mc_cr_norm_dff_ti2.hdf5" to memory...
  27-Aug-2015 18:34:55: Done!
  27-Aug-2015 18:35:01: Movie will be displayed with fixed CLim = [-0.038 0.051]...
  27-Aug-2015 18:35:08: Loaded data from sss1_new\rec_150826-223018.mat
  27-Aug-2015 18:35:08: No classification file in sss1_new!
  27-Aug-2015 18:35:28: Loaded filters/traces from "sss1_new"
Classifier (1/911) >> m
  Showing cell map (press any key to return)
```
which shows the following figure:
image

The current cell is shown in yellow. Unlabeled cells are shown in white. 

As the classification proceeds, classified cells are shown in green, and rejected cells shown in red. For example:
```
Classifier (1/911) >> p!
  Candidate 1 classified as phase-sensitive cell
Classifier (2/911) >> p!
  Candidate 2 classified as phase-sensitive cell
Classifier (3/911) >> m
  Showing cell map (press any key to return)
```
yields:
image
(To be honest, I think the white is a bit overwhelming, and the yellow outline doesn't sufficiently pop out. I propose to tweak the display settings this over time.)

You can also click on a cell boundary, to get its index via tooltip:
image

The idea was that one may want to "hop around" the cell indices during cell classification, for which such a "map" tool is helpful.


forea:
based on quick test, loads more quickly and 'm' map function works

Re: yellow outline doesn't sufficiently pop out
how about filling in the boundaries? or, as you mentioned, thicker line width

kimtonyhyun:
Let me play around with line thicknesses...

inanhkn:
Does the cell ID display only work when we click the boundary? How about when clicking on an interior point?

kimtonyhyun:
In principle possible, but would take more effort. (At that level of effort, I think there are even better -- non-tooltip based -- approaches.) I propose to defer the interior click.

kimtonyhyun:
@inanhkn Challenge accepted.

classification/support/view_cell_interactively.m
```diff
@@ -260,7 +260,7 @@ function show_other_cells(show)
         
         for m = 1:num_other_cells
             set(other_cell_handles(m,1), 'Visible', vis_val);
-            set(other_cell_handles(m,2), 'Visible', vis_val);
+%             set(other_cell_handles(m,2), 'Visible', vis_val);
```
inanhkn: How about running the `text()` command here so we can still display the cell indices? If an average user does not use the neighbors functionality too often, the neighbors showing up late should not be a big issue.

kimtonyhyun:
You can now click on the interior of boundaries. Additionally, the interiors are colored, which makes it easier to visualize (e.g. where is the yellow cell?):
image

inanhkn:
How about we make the current cell purple or some other color with high contrast with respect to white?

classification/support/view_cell_interactively.m
kimtonyhyun: Yes, I also thought of the lazy eval approach. However, there's still the issue that 1000 `text`s on the small movie is actually useless.

I think the right approach is to use tooltips here as well, or programmatically determine when the plot view is over a particular ROI of the movie, and apply `text`s over local regions. The latter approach would be significantly more work.

kimtonyhyun:
Magenta is no good because it has poor contrast against red. Tweaked border thicknesses and transparencies, and using cyan for contrast:
image

classification/support/view_cell_interactively.m
kimtonyhyun: In fact, I would propose a tooltip approach here too. It's way more functional.

kimtonyhyun: In fact, turning off the `text` here makes the "n" option way more useful.

kimtonyhyun:
Btw, the "m" toggle makes use of the `plot_cell_map` method of `DaySummary`, so the colored plotting of cellmaps can be used outside of `classify_cells.m` as well, i.e.:
```
ds = DaySummary(...);
ds.plot_cell_map();
```

kimtonyhyun:
In the latest commit, I've extracted `DaySummary.m` into a top-level directory `ds`. The reason is that I believe that the `DaySummary` object is a very natural building block for further analyses, and I have a  number of ideas that operates at the level of `DaySummary`s.

Please note that you'll have to add `ds` (and its subdirectories) to your paths.

kimtonyhyun:
I've now explicitly removed the `text`-based cell indices from `view_cell_interactively.m`. Instead, I added coloring for the "other cell boundaries" that is consistent with the cell map: e.g. cyan is the current cell, green is classified cells, red is rejected cells, white is unlabeled:
image

kimtonyhyun:
@forea If the map function is working for you, can you merge -- as this PR is the basis for a number of further `DaySummary` work?

PR #119, opened 2015-08-29
Inanhkn/pca sss

inanhkn:
I added functionality in `extract_cells` that identifies and merges the 'split' cells. For that the script looks at the traces belonging to the cells nearby and decides they are a part of a same cell if the correlation between the (normalized) traces are higher than 0.65 (1 is maximum possible). There is no science to the value 0.65, but trial and error approach suggested it is a reasonable value. We can surely modify it later on. 

I also made the trace extraction a standalone function named `extract_traces` that takes movie source and rec source as inputs. Both movie source and rec source might be either the path to the source or the source itself. For example both of the following are valid calls to `extract_traces`:

```
1)
traces = extract_traces('dum.hdf5','dum.rec');

2)
M = [movie  matrix];
F = [filters matrix];
traces = extract_traces(M,F);

```
movie matrix can be fed as 2D or 3D; both are accepted.

Lastly,  I added 2 scripts in the `/utils` directory, namely `dfs.m` and  `sparse_to_csr.m`, which are necessary for split cell merging. I did not check inside the scripts; all I know is that they do depth first search. I just did not want to write my own version of DFS.


inanhkn:
For testing this version on c11m1d12, you can use the following PCA file which has 807 PCs processed with the median filter with halfwidth of 2. This is the same file that was used for the rec*.mat that yielded the above results.
https://www.dropbox.com/s/5vlq2dcbc9tg3ta/pca_n807.mat?dl=0

signal_op/extract_traces.m
```diff
@@ -0,0 +1,53 @@
...
+        else
+            error('Sizes of movie matrix and filters must be consistent.')
+        end
+    else
+        num_filters = size(rec_source,3);
+        F = reshape(rec_source,h*w,num_filters);
+    end
+end
+
+idx_nonzero = find(sum(F,2)>0);
```
kimtonyhyun: My understanding is that, here, subselecting by `idx_nonzero` or not would actually yield the same traces at the end, correct? This is purely a computational optimization with no effect on the result?

inanhkn:
That's right.

inanhkn:
Here's a dropbox link to the rec file that came out of this script processed with c11m1d12 movie and the pca file I shared above:
https://www.dropbox.com/s/93p5w8hrum9arbq/rec_150828-221528.mat?dl=0

kimtonyhyun:
I'll be able to take a look at this in the evening.

kimtonyhyun:
For reference, can you start providing explicit runtimes on the PCA and extraction (and the name of the PC that ran it)?

inanhkn:
I'm sharing an example console output from `extract_cells`:

```
29-Aug-2015 14:21:21: Loading movie singular vectors...
29-Aug-2015 14:21:24: Extracting filters...
	 	 	 Using GPU to accelarate processing
	 	 	 Extracted 1000 potential cells 
29-Aug-2015 14:22:11: Cleaning filters and removing duplicates...
29-Aug-2015 14:22:35: Loading movie for trace extraction...
29-Aug-2015 14:22:38: Extracting traces...
29-Aug-2015 14:23:09: Identifying possible split cells...
			 Identified 62 split cells (total of 167 objects), merging now...
29-Aug-2015 14:23:12: Extracting traces with updated filters...
29-Aug-2015 14:23:25: Cell Statistics:
 			 Total of 895 potential cells are extracted.
 			 First 411 extracted cells are likely to be cells.
 			 It is likely that there are few or no cells from  #856 onwards. 
29-Aug-2015 14:23:25: Writing the output to file... 
29-Aug-2015 14:23:25: Output will be saved in rec_150829-142325.mat 
29-Aug-2015 14:23:32: All done! 
```
This instance took 2 minutes and 11 seconds.

Here's the link to the rec file generated (c11m1d12):
https://www.dropbox.com/s/6tm6z6gyce9o9g6/rec_150829-142325.mat?dl=0

inanhkn:
As a note, I feel that finding so called 'split cells' is not merely a quirk of the source separation algorithm; some(or many) of the split looking cells might actually be individual cells. In many cases it is not apparent to the human eye as SNR might be low and there might be cells overlapping in x,y but distinct in the z dimension.

kimtonyhyun:
If they are distinct cells, then their temporal correlation should be low. Can you quantify?

-Tony (from phone)

inanhkn:
You're referring to a somewhat ideal scenario where the activities are isolated enough. This is not interesting because we would not talk about a "split-looking cell" to begin with in that case. I'm referring to a situation where for instance an "oblong" cell dominates nearby region and interferes with the activities of other cells with low SNR. I have come across cells like this when trying to improve the cell merging; I'll try to add pictures as I see other examples (I'm guessing adding pictures is what you mean by 'quantify'.).

kimtonyhyun:
"Quantify" means support your case with evidence, whether it is pictures or whatever,

-Tony (from phone)

kimtonyhyun:
So far, through the first 100 cells (of 895).

There are interesting examples where the filter merging helps (a particularly striking one is cell index 63):
image

There also others where filters were merged for no reason (i.e. one good filter and one neighboring irrelevant). Example is cell 58, where the top lobe is a cell, and bottom lobe is nothing:
image

I had two thoughts at this point:
1. We could let the user, at the time of classification, decide to merge filters or not. It would take some coding for the UI, but once in place, this would result in much smarter filter merging.
2. As an alternative approach to merging of split cells (which yields irregular-shaped cells, like 63 above), is there a way to relax the "tightness" of cells so that a single filter would capture the large, oblong cells?

kimtonyhyun:
If we're considering the possibility of interactively extracting cells (e.g. choice of merge), one might also allow the user to add in ROIs during classification, if a cell is observed in the movie that's not captured by an existing boundary.

inanhkn:
We could implement features like "merging", creating new ROIs etc. in `classify_cells`, yes, We could also implement applying the previously classified cell map on the current movie and auto-classifying the matching cells (you mentioned you started doing this recently). 

As for the source extraction algorithm, it does not have a sense of "tightness" of a source; it's a generic algorithm that extracts whatever it thinks is a source.

kimtonyhyun:
Yeah, the suggestion was to think of potential alternatives of grabbing the oblong cells, rather than merging multiple circular bases together -- more detailed report sometime today.

kimtonyhyun:
_Final result: 413 of 895_
image

Comparison to previous SSS run which had 358 of 734 (previous run in blue, latest in red):
image

Comparison to the original CM run, which had 424 of 888 (CM run in blue, latest SSS in red):
image

This is getting interesting. My overall observations are that:
- CellMax performs better with oblong shape near the left edge of the field of view. CM filter shape matches the underlying excitation better in general.
- SSS seems to have higher sensitivity for cells that spike only a few times. (Note that SSS found many cells not in the "record" CM run.)

I have more detailed feedback on SSS handling of split cells, to come later.

I also think it would be worth our time to develop a mechanism where we can append filters to an existing filter set (whether it's runtime merging, runtime addition of ROI, etc.). That way, cell extraction result for a given movie is guaranteed to get monotonically better as we work with it, i.e. it's not an all-or-nothing proposition of taking this extraction result or the other.

inanhkn:
This is interesting. I actually realized I had a small knob that I can tweak to get probably more cells. The algorithm iteratively extracts sources on the condition that each new source has less than X (normalized to [0,1]) correlation to all the previously extracted cells. This was 0.1 until yesterday, and I decided to increase it to 0.2. I can see that in the latest run SSS was able to find more spatially overlapping cells due to this. Increasing too much is against the philosopy of the algorithm since ideally we want the sources(cell shapes) as independent as possible(hence as orthogonal as possible). Contrary to intuition, it is possible to have orthogonal filters that are highly overlapping, but it is not as likely as if they were not overlapping.

As another note, I have a hard limit(1000) on the number of sources to be extracted, I can increase it to see if more cells come out of it. I can also try further increasing the X value mentioned above.

Tony, could you comment on the quality of sorting of the cells? Or better yet, could you add a plot that shows accumulative classified cells against the potential cell index?

kimtonyhyun:
Yeah, I can give some reports on the sort quality too. At a high level, it's pretty good, but there are definitely very strong exceptions.

inanhkn:
Yes, I expect that there are very good cells beyond the trust region of the sorting method, but I expect that there are really few obviously bad cells inside the trust region(411 in this run).

kimtonyhyun:
Label in order of sort order:
image

I revised some labels. The count is now __417 of 895__.

First three rejected cells:
- 157 is a split cell
- 181 is a blood vessel, with zero signal
- 186 is a split cell (borderline acceptable, in this case)
- 210 is a split cell
- 214 is a split cell

Classification file for SSS3: https://www.dropbox.com/s/ivqiolouzs7r5cl/class_150830-141756.txt?dl=0

inanhkn:
By the way, it is interesting to see cells captured in SSS2 and not SSS3 that look very isolated. I am fairly confident that this is due to the hard limit we have since those cells have almost zero spatial correlation with all the remaining candidates, and the algorithm tries to extract everything present in the PCs (same PCA file was used for both runs). I'll increase the limit.

inanhkn:
Thanks for the plot Tony, I'll try to better correlate my sorting criterion to your classification results and come up with better sorting in the future.

inanhkn:
Alright. I increased the limit to 1500, and the algorithm found 1295 cells. Here's the console output from this run:

```
30-Aug-2015 14:45:32: Loading movie singular vectors...
30-Aug-2015 14:45:35: Extracting filters...
	 	 	 Using GPU to accelarate processing
	 	 	 Extracted 1500 potential cells 
30-Aug-2015 14:46:32: Cleaning filters and removing duplicates...
	 	 	 1 cells were eliminated due to minimum size requirement(<10 pixels)
30-Aug-2015 14:47:15: Loading movie for trace extraction...
30-Aug-2015 14:47:18: Extracting traces...
30-Aug-2015 14:48:03: Identifying possible split cells...
			 Identified 108 split cells (total of 312 objects), merging now...
30-Aug-2015 14:48:09: Extracting traces with updated filters...
30-Aug-2015 14:48:27: Cell Statistics:
 			 Total of 1295 potential cells are extracted.
 			 First 458 extracted cells are likely to be cells.
 			 It is likely that there are few or no cells from  #591 onwards. 
30-Aug-2015 14:48:27: Writing the output to file... 
30-Aug-2015 14:48:27: Output will be saved in rec_150830-144827.mat 
30-Aug-2015 14:48:36: All done! 
```

This one took 3 minutes and 4 seconds. Below is the rec file generated:
https://www.dropbox.com/s/rzljyaakv6b08i6/rec_150830-144827.mat?dl=0

Adopting @kimtonyhyun 's convention, I call the latest rec result SSS4. Here's the comparison of all found cells of SSS4 (red) to classified cells of SSS3(blue): (Cells of SSS3 are taken from @kimtonyhyun 's classification result)
sss3_sss4

Apart from the merged cells, the overlapping cells look almost identical; @kimtonyhyun you can perhaps auto-label the overlapping cells and classify the remaining cells of SSS4?

kimtonyhyun:
I suggest that you improve the handling of the oblong cells in the left side of the FOV, rather than simply ramping up the total cell candidate counts.

Here is the "cm01" reference run (424 of 888) rec file and classification results. Containing folder also has all extraction runs (CM, SSS) that I've done so far: https://www.dropbox.com/sh/13xmit6iyrl5izr/AADkPMnAZh6U_SuCFdSi01o_a?dl=0

inanhkn:
Although I agree that the handling of the oblong cells is an issue to work on, I believe there is always value in trying to extract more cells, regardless of shape. I will of course spend time on the present issues; in the meanwhile I thought it would be adequate to ask you to classify the latest result since 1) most of the cells are overlappin with the previous run and the new run is able to capture almost 100% of the previously classified cells and potentially more good cells 2) I believe you classifying as opposed to me doing it would be more unbiased since all the previous results belong to you.

kimtonyhyun:
Ok.

kimtonyhyun:
These notes regarding split cells and cell merging pertain to SSS3.

In one case, I observed two nearby but distinctly firing cells, to be incorrectly merged as one. ID: 150. The two cells are high-reliability cells (fires almost every trial) and the two are displaced only a little in time while being spatially adjacent.

In some other cases, I observed two filters merged into one, where one is a cell and the other is not. Presumably, this is innocuous in terms of the extracted trace, but the filter shapes are incorrect which may affect cross-day alignment performance among others. IDs: 17, 18, 58, 423

I also noted the instances where the cell merging resulted in better capture of the cell in the movie (at least for trace extraction purposes). IDs: 23, 26, 52, 63, 72, 77, 121, 122, 177, 196, 207, 208, 373, 417, 430, 440, 481, 567.

However, the merged filters generally do not represent the underlying cell shape well. They look like a cover of the cell using circular basis functions. (Some exceptional cases reported below.) Hence, I would advocate for an alternative approach (non-merge) that captures the oblong cells.

Cell 63:
image

Cell 121:
image

Cell 196:
image

Cell 207:
image

Cell 208:
image

Cell 481:
image

kimtonyhyun:
Or, when it's decided that a particular set of filters need to be merged, a more sophisticated approach (than the direct sum) may be better, e.g. some sort of blending, or trigger a re-extraction of the filter with minimum size requirements.

kimtonyhyun:
Latest classification in SSS4: __445 of 1295__

I copied the labels from SSS3, as shown here:
sss4_labels-copied

I then classified the nonmatched cells:
sss4_labeled

Here's the comparison showing only classified cells between SSS3 (blue) and SSS4 (red):
sss4_vs_sss3

Here's the comparison against CM01 (in blue; 424 of 888) vs. SSS4 (red):
sss4_vs_cm01

My impression is that SSS is particularly strong at finding sparse cells -- just a few calcium events in the entire movie -- as long as they are circular.

Let's improve the handling of oblong cells, and I think the SSS approach would be very strong in terms of completeness of extraction and obviously speed. (You should also cite the time taken for PCA computation -- not just `extract_cells`.)

I uploaded my classification results to: https://www.dropbox.com/s/v5srpkizn0szrb5/class_150831-100308.txt?dl=0

Can you take the classified cells only from SSS4 and re-extract traces using only the classified subset?

inanhkn:
Not to nitpick or anything, but I'd like to share the map of all the cells found by SSS4(red) versus classified cells of CM01(blue) :
cm_vs_sss4

Although SSS4 obviously has issues with oblong cells, it seems to capture almost everything that CM01 does. Below is some example cells that were not classified as cell in SSS4 but classified as cell in CM01. each row represents one area, first column is taken from the figure right above(all cells of SSS4 vs classified CM01), and the second column is taken from the figure in @kimtonyhyun 's post above (classified cells of SSS4 vs classified cells of CM01) : (The very poorly drawn arrows are meant to point to the cells of interest)
cm_vs_sss4_zoom

I will think about what can be done to capture better the elongated cells.

inanhkn:
@kimtonyhyun  I will update the `extract_traces` routine to take in an optional classification file input to do what you are asking.

inanhkn:
By the way, the PCA run that gave the mat file used in all the SSS runs took my computer about 22 minutes (roughly 10 minutes was for median filtering with halfwidth of 2).

kimtonyhyun:
For fun, here's a comparison between SSS4 and "Rec1" -- the initial extractions we were using as we were taking a look at the data "on the fly":
image

At least we're making some progress.

Wait a minute, there is a cell in Rec1 that's not captured by SSS4. It's an awesome one too:
image

kimtonyhyun:
Here's the comparison between Rec1 and CM01:
image

Rec1 also got one over CM01:
image

kimtonyhyun:
Merging to play with `extract_traces`

inanhkn:
What is it exactly ou plan to do with `extract_traces`? Should I not work on the update to incorporate classification into it?

signal_op/extract_traces.m
```diff
@@ -0,0 +1,53 @@
+function traces = extract_traces(movie_source,rec_source)
+% Extract traces from movie source and rec source. The inputs can be either
+% the paths to the sources or the sources themselves. Specifically, movie
+% can be fed as the path, 3D movie matrix or the 2D (pixels flattened)
+% movie matrix. Likewise, the rec source might be the path to the rec file
+% or the 'filters' structure in the rec file (or its transpose).
+
+regul = 0; % regularization term for LS reconstruction
```
kimtonyhyun: On second thought, shouldn't we use some sort of nonzero regularization here? My understanding is that when filters are nonorthogonal, then we have in principle multiple solutions to the LS problem, and we need to select a particular solution by additional constraints -- e.g. regularization.

kimtonyhyun: Actually, I'm thinking of _nonindependent_ filters, not _nonorthogonal_.

PR #121, opened 2015-08-31
Kimth/classify ds

kimtonyhyun:
Refactoring of `DaySummary` and `classify_cells`. Most changes are behind the scenes. No additions to Matlab path are required.

The main conceptual change is that `classify_cells` is thought of as an operation on an existing `DaySummary`. The proper way to invoke `classify_cells` is now as follows:
```
>> sources = data_sources

sources = 

          pca: '_data/pca_n700.mat'
          fps: 10

>> M = load_movie(sources.miniscope);
>> m1d12_cm01 = DaySummary(sources.maze, 'cm01');
  31-Aug-2015 12:15:14: Loaded trial metadata from _data/c11m1d12_ti2.txt
  31-Aug-2015 12:15:18: Loaded filters/traces from cm01\rec_150819-203737.mat
>> classify_cells(m1d12_cm01, M);
```
Note that `classify_cells` now expects a movie matrix directly. Note also that the use of the `sources` structure is less motivated than before. (I still use it for convenience.)

The conceptual change where `classify_cells` "operates" on a DaySummary was motivated by the workflow where one can copy labels from one DaySummary to another (with sufficient spatial overlap), and then `classify_cells` is used to label the unmatched cells.

Within `classify_cells`, there is one new behavior which is that, when you press "[Enter]" in the prompt, the routine will jump to the next unlabeled cell (in index).

### DIFF
--- a/alignment/bijective_filter.m
+++ b/alignment/bijective_filter.m
@@ -1,0 +1,37 @@
+function [match_1to2, match_2to1] = bijective_filter(match_1to2, match_2to1)
+
+for i = 1:length(match_1to2)
+    match_itoj = match_1to2{i};
+    num_matches = size(match_itoj,1);
+    if (num_matches > 1)
+        for k = 2:num_matches
+            j = match_itoj(k,1);
+            match_jtoi = match_2to1{j};
+            [~, diff_inds] = setdiff(match_jtoi(:,1), i);
+            if ~isempty(diff_inds)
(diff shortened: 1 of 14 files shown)


kimtonyhyun:
The new organization of `DaySummary` uses class folders. Look here for more info: http://www.mathworks.com/help/matlab/matlab_oop/class-files-and-folders.html

forea:
works for me, merging.

PR #126, opened 2015-08-31
Kimth/refactor alignment

kimtonyhyun:
Slight refactor of `run_alignment.m`, which was written before the use of `DaySummary`. (It was quickly retrofitted to make use of `DaySummary` inputs.)

No changes to the overt behavior of `run_alignment.m`. Alignment script can be used as before. The boot-up process for `run_alignment.m` should be a bit faster now, since it makes use of pre-computed variables within `DaySummary` (i.e. boundaries and masks, which already exists within `DaySummary`).

In general, I am absorbing various functionalities (e.g. the ability to click on a cell by mouse click) into `DaySummary`. This allows us to use the functionalities in other scripts -- my eye is set on merging filters from multiple `DaySummary`s.


forea:
Tested and works for me. Merging now.

PR #132, opened 2015-09-03
Kimth/make sbs

kimtonyhyun:
The function `make_sidebyside.m` will generate an AVI file that has the behavior and Miniscope recordings laid out horizontally:
image

The function can be invoked as follows:
```
>> make_sidebyside('c11m5d10.txt', 'c11m5d10.mp4', 'c11m5d10_gfix_cr_rm_mc_cr_norm_dff.hdf5', [35 45 66]);
```
where the last input (`[35 45 66]` in this case) indicates the trials to be saved to the AVI file.

The behavior video (MP4 file), the miniscope movie (HDF5 file) and the frame indices (TXT file) must all have matching number of frames, otherwise the script will abort. Hence, note that it is possible that the Ubuntu systems may not be able to run this script -- given the issues we've had in the past with Ubuntu Matlab saying that there is one missing frame in the behavior MP4 files.

Note that the script loads chunks of the Miniscope movie from disk (in units of trials), so that the script is compatible with machines that don't have large amounts of ram (e.g. 32+ GB).

The output AVI file is compatible with Windows Media Player, VLC Player, Quicktime on my home machine (as well as "reuniens"), as well as Dropbox preview. @forea: Can you check on your PCs as well? An example output is here: https://www.dropbox.com/s/gaorv04w9groiev/c11m5d10_sbs.avi?dl=0

### DIFF
--- a/view/make_sidebyside.m
+++ b/view/make_sidebyside.m
@@ -1,4 +1,14 @@
 function make_sidebyside(maze_source, behavior_source, miniscope_source, trials_to_record)
+% Generate a side-by-side AVI file of the specified trials.
+%
+% Example usage:
+%   make_sidebyside('c11m5d10.txt',...
+%                   'c11m5d10.mp4',...
+%                   'c11m5d10_gfix_cr_rm_mc_cr_norm_dff.hdf5',...
+%                   [35 45 66]);
+%
+% NOTE: `miniscope_source` must be a HDF5 file
+%
 
 % Get the table of frame indices
 %------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


kimtonyhyun:
In the future, we could add features like:
- Overlay "Trial X" on the behavior movie
- Make the side-by-sides vertically, as opposed to horizontally

forea:
output AVI file looks fine on my PC. merging

(as expected, does not run on Ubuntu)
Error using make_sidebyside (line 25)
Number of frames in behavior video (27760) does not match that in the maze file (27761)!

PR #134, opened 2015-09-04
Kimth/load behavior

kimtonyhyun:
@forea, @inanhkn This is an initial foray into integrating the behavior video into our analysis workflow. My guess is that things will evolve a lot from this point (e.g. visualization methods), but wanted to run by the initial implementation.

First, you should instantiate a `DaySummary` object as before:
```
>> sources = data_sources

sources = 

     behavior: '_data/c11m1d12_ti2.mp4'
          pca: '_data/pca_n700.mat'
          fps: 10

>> m1d12_cm001 = DaySummary(sources.maze, 'cm001');
03-Sep-2015 16:12:40: Loaded trial metadata from _data/c11m1d12_ti2.txt
03-Sep-2015 16:12:43: Loaded filters and traces from cm001\rec_150817-165436.mat
03-Sep-2015 16:13:01: Loaded classification from cm001\class_150819-123923.txt
```

By default, `DaySummary` does not initialize the behavior video. Instead, you can "load" the behavior video into `DaySummary` afterwards as follows:
```
>> m1d12_cm001.load_behavior_movie(sources.behavior);
03-Sep-2015 16:20:07: Loaded behavior video from "_data/c11m1d12_ti2.mp4"
```
For now, I made the behavior video something you can optionally load after the initial construction of `DaySummary`. I did it this way for now, because I thought that there may be cases (e.g. classification of cells) where we would use `DaySummary`s without having to refer to the behavior video. We could use alternate loading schemes in the future.

The use of the `sources` struct is optional. You can directly specify the path to the behavior video. Note: `load_behavior_movie` will issue a warning if the number of frames in the behavior movie does not match the expected number of frames in the frame indices table. This could be an issue on Ubuntu systems where the behavior videos (MP4 files) seem to underreport by one frame.

As a start, I created one "application" that makes use of the behavior video (`browse_rasters.m` which I introduced a little bit ago). Using the "c11m1d12_cm001" dataset:
```
>> browse_rasters(m1d12_cm001);
Raster browser (page 1 of 53) >> 
```
image

I then select a particular cell, to view the detailed raster. For example, cell 23:
```
>> browse_rasters(m1d12_cm001);
Raster browser (page 1 of 53) >> 23
  Cell 23 selected!
Cell 23 raster >> 
```
image

At this point, you can type '3' to look at trial 3:
image
which shows the trace for this cell in trial 3, and the corresponding behavior video. You are now able to click on the top right trace, to view the corresponding behavior frame:
image

You can return to the raster by hitting any key on the keyboard:
```
Cell 23 raster >> 3
  Showing trial 3. Press any key to return.
Cell 23 raster >> 
```

This is just the first pass at integrating the behavior video -- I'm sure we'll come up with better visualizations as we further work with behavior videos.

### DIFF
--- a/ds/@DaySummary/DaySummary.m
+++ b/ds/@DaySummary/DaySummary.m
@@ -261,7 +261,8 @@ classdef DaySummary < handle
         %------------------------------------------------------------
         function load_behavior_movie(obj, behavior_source)
             obj.behavior_vid = VideoReader(behavior_source);
-            fprintf('%s: Loaded behavior video from "%s"\n', behavior_source);
+            fprintf('%s: Loaded behavior video from "%s"\n',...
+                datestr(now), behavior_source);
             if (obj.behavior_vid.NumberOfFrames ~= obj.trial_indices(end,end))
                 fprintf('  Warning! Number of frames in behavior video (%d) does not match the trial frame table (%d)!\n',...
                     obj.behavior_vid.NumberOfFrames, obj.trial_indices(end,end));
--- a/ds/@DaySummary/DaySummary.m
+++ b/ds/@DaySummary/DaySummary.m
@@ -259,6 +259,10 @@ classdef DaySummary < handle
         
         % Load behavior movie
         %------------------------------------------------------------
+        function loaded = is_behavior_loaded(obj)
+            loaded = ~isempty(obj.behavior_vid);
+        end
+        
         function load_behavior_movie(obj, behavior_source)
             obj.behavior_vid = VideoReader(behavior_source);
             fprintf('%s: Loaded behavior video from "%s"\n',...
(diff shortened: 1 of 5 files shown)


kimtonyhyun:
Btw, `freezeColors.m` and `unfreezeColors.m` are just Matlab Exchange functions that allow us to use multiple colormaps in the same subplot.

kimtonyhyun:
I think the current tool is mostly useful for visualizing where the "trial phase" maps to in real positions. For example:
image

kimtonyhyun:
I also put a time-binned version of the c11m1d12 behavior movie here, in case you need it for trying out the tool: https://www.dropbox.com/s/bmpi2jacq7bpsme/c11m1d12_ti2.mp4?dl=0

